PR #2, opened 2025-01-14
Support content + tool calls

WonderPG:


BoBer78 (approved):
LGTM !

jankrepl:
as discussed, I am not against it but not sure if this is high priority. However, if we go for it I would also make sure we can somehow trigger when we want and make sure it does not break the frontend logic. As per our agreement, if a backend change has a potential to break something in the frontend we should also adjust the frontend right away in the PR

PR #3, opened 2025-01-15
Add pagination nd thread metadata endpoint

### LINKED ISSUE #5: Add pagination to threads, messages and tool calls

WonderPG:
Closes #5

### DIFF
--- a/src/neuroagent/app/database/schemas.py
+++ b/src/neuroagent/app/database/schemas.py
@@ -3,7 +3,7 @@
 import datetime
 from typing import Any, Literal
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from neuroagent.app.database.sql_schemas import Entity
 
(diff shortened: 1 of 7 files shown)


src/neuroagent/app/database/schemas.py
```diff
@@ -42,3 +42,36 @@ class ToolCallSchema(BaseModel):
...
+    """Paginated output for ThreadRead endpoints."""
+
+    results: list[ThreadsRead]
+
+
+class PaginatedMessagesRead(PaginatedOutputBase):
+    """Paginated output for ThreadRead endpoints."""
+
+    results: list[MessagesRead]
+
```
jankrepl, line 72: You can do this entire thing in a "cleaner" way using generic types.

But the current approach is probably simpler to understand. No actions neede

WonderPG: Ah sure using one class which has a `results: list[BaseModel]` right ?

jankrepl (dismissed):
Great:))) Thanks!

src/neuroagent/app/database/schemas.py
jankrepl: It would something along the lines of 


```python
T = TypeVar("T", bound=BaseRead)


class PaginatedOutput(BaseModel, Generic[T]):
    """Pagination schema."""
    page: int
    page_size: int
    total_pages: int
    results: list[T]
```


But no need to implement it...probably might be confusing more than helpful

jankrepl (approved):

BoBer78 (approved):

jankrepl:
Can we also do pagination on the messages?

jankrepl:
Closing. We might revisit at some point but from scratch.

PR #58, opened 2025-01-27
Make human message max width smaller

### LINKED ISSUE #34: Human message should never make it all the way left
Probably set the maximum width of the human message to something lower than 100%

Image

Code: 

https://github.com/openbraininstitute/neuroagent/blob/cf9e19530fbe02106749673c37592392b3c8077f/frontend/src/components/chat-message-human.tsx#L11C1-L19C11

I guess a combination of https://tailwindcss.com/docs/max-width#using-a-percentage and `ml-auto`. Also, we can drop the inner `<div>`

jankrepl:
Closes #34


BoBer78 (approved):
👍

PR #64, opened 2025-01-28
Collapse side bar

### LINKED ISSUE #50: Make sidebar (with threads) collapsible
ChatGPT has that feature. Probably nice if you want to present the current thread messages without "leaking" all your past conversations.

Note that there is already an icon that one would click but nothing is hooked up to it yet.

Image

BoBer78:
Closes #50

### DIFF
--- a/frontend/src/components/sidebar.tsx
+++ b/frontend/src/components/sidebar.tsx
@@ -1,46 +1,66 @@
 "use client";
 
 import { useState } from "react";
-import { PanelRightOpen, Search, SquarePen } from "lucide-react";
+import {
+  PanelRightOpen,
+  PanelRightClose,
+  Search,
+  SquarePen,
+} from "lucide-react";
 import Link from "next/link";
+import { usePathname } from "next/navigation";
 
 export function Sidebar({ children }: { children: React.ReactNode }) {
+  const pathname = usePathname();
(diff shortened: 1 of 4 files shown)


frontend/src/components/sidebar.tsx
```diff
@@ -1,25 +1,46 @@
...
+          <div className="overflow-auto flex-1">{children}</div>
         </div>
-        <div className="flex p-4 gap-4 ">
-          <Search className="hover:scale-[1.1] transition" size={24} />
-          <Link href="/">
-            <SquarePen className="hover:scale-[1.1] transition" size={24} />
-          </Link>
+      ) : (
+        <div className="p-4 border-r-2">
+          <PanelRightOpen
```
jankrepl, line 37: Here I would put https://lucide.dev/icons/panel-right-close

jankrepl:
> Not sure if this is enough ? should we maybe auto-collapse the sidebar when we are not on a threads page ?

@BoBer78 I think the current functionality is perfect:) I would not overcomplicated it:)

BoBer78:
I added a slide transition, the button, and the fact that it is disabled on the login screen

jankrepl:
Amazing and simple:) Feel free to merge:)

PR #67, opened 2025-01-29
Confirmation dialog before thread deletion

### LINKED ISSUE #40: Add confirmation dialog to thread deletion
Currently, clicking the `X` icon will right away delete the thread. Instead, we should have a dialog (basically the same dialog as the edit) that asks the user to confirm.

https://ui.shadcn.com/docs/components/dialog

Relevant code: https://github.com/openbraininstitute/neuroagent/blob/main/frontend/src/components/thread-card-sidebar.tsx

WonderPG:
Closes #40


jankrepl:
So the `thread-card-sidebar.tsx` is getting very heavy. Could I suggest creating 2 new components

* `src/components/edit-thread-dialog.tsx` `EditThreadDialog`
* `src/components/delete-thread-dialog.tsx` `DeleteThreadDialog`


and just import them + render them inside `thread-card-sidebar.tsx`?

@WonderPG

WonderPG:
Done in e15037d

frontend/src/components/delete-thread-dialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <DialogTitle>Delete this thread ?</DialogTitle>
+        </DialogHeader>
+        <DialogFooter>
+          <Button
+            type="submit"
+            onClick={() => {
+              setIsDialogOpen(false);
+            }}
+          >
+            Cancel
```
jankrepl, line 55: I find the `Cancel` button redundant since by default there is the `x` in the top right corner anyway.

Also, the edit dialog does not have it either.

WonderPG: But the HIL validation has it 
image
We should stick to one way of doing it

BoBer78:
Small UI comment : I would out the delete button in red, or something to signal the user that the button is "important".

WonderPG:
I do like the full greyscale UI we have so far. Adding a red button would be the first bit of color we have. What do you think?

frontend/src/components/delete-thread-dialog.tsx
WonderPG: We decided to keep it and remove the `x` top right

jankrepl (approved):
LGTM:)

BoBer78 (approved):
LGTM 🏖️

PR #69, opened 2025-01-30
[Frontend] - fix delay on first message 

### LINKED ISSUE #41: Starting a new thread (by typing a question) too slow
The problem is that currently, we trigger the following action

https://github.com/openbraininstitute/neuroagent/blob/main/frontend/src/actions/create-thread.ts

It does does the following things

1. Hits create thread
2. Hits chat streamed and waits until the stream is finished
3. Redirect to `/threads/{thread_id}`

Probably a better approach would be 
1. Somehow store the user message
2. Hit create thread
3. Redirect to `/threads/{thread_id}`
4. Insert the stored initial user message to `useChat`

BoBer78:
Closes #41


jankrepl (approved):
LGTM:)

Issue #44, opened 2025-01-24
List tools endpoint


jankrepl:
! Attention, this is not about listing things from our DB (e.g. things in tool-calls or messages). This is about the "metadata" of available tools

For each tool it would be nice to see 

* name
* description
* input parameters

I guess there we don't really need to add pagination. Maybe in the future we could also do something similar for different agents (what tools are part of what agent) but whatever.

Anyway, the idea is that in the frontend one will be easily able to nicely list all the tools they can use

PR #91, opened 2025-02-06
Make it possible for user to chose tools dynamically

### LINKED ISSUE #84: POC allow user to select which tools should be enabled
The idea would be to have a button in the frontend that allows the user to "uncheck" certain tools so that the agent uses only the ones selected.

Could be used jointly with #44

WonderPG:
Closes #84

### DIFF
--- a/backend/src/neuroagent/app/routers/tools.py
+++ b/backend/src/neuroagent/app/routers/tools.py
@@ -105,7 +105,7 @@ async def execute_tool_call(
 
 
 @router.get("")
-def get_tool_list(
+def get_available_tools(
     tool_list: Annotated[list[type[BaseTool]], Depends(get_tool_list)],
 ) -> list[str]:
     """Return the list of available tools."""
--- a/backend/src/neuroagent/app/routers/tools.py
+++ b/backend/src/neuroagent/app/routers/tools.py
@@ -17,6 +17,7 @@ from neuroagent.app.dependencies import (
     get_session,
     get_thread,
     get_tool_list,
+    get_user_id,
 )
 from neuroagent.app.schemas import (
     ExecuteToolCallRequest,
@@ -107,6 +108,7 @@ async def execute_tool_call(
 @router.get("")
 def get_available_tools(
     tool_list: Annotated[list[type[BaseTool]], Depends(get_tool_list)],
+    _: Annotated[str, Depends(get_user_id)],
 ) -> list[str]:
     """Return the list of available tools."""
     # Trivial implementation for now, to be adressed in another PR
--- a/backend/src/neuroagent/app/routers/tools.py
+++ b/backend/src/neuroagent/app/routers/tools.py
@@ -15,14 +15,14 @@ from neuroagent.app.dependencies import (
     get_agents_routine,
     get_context_variables,
     get_session,
-    get_starting_agent,
     get_thread,
+    get_tool_list,
 )
 from neuroagent.app.schemas import (
     ExecuteToolCallRequest,
     ExecuteToolCallResponse,
 )
-from neuroagent.new_types import Agent
+from neuroagent.tools.base_tool import BaseTool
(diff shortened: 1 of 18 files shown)


backend/src/neuroagent/app/routers/tools.py
```diff
@@ -102,3 +102,12 @@ async def execute_tool_call(
     await session.commit()
 
     return ExecuteToolCallResponse(status="done", content=message["content"])
+
+
+@router.get("")
+def get_available_tools(
+    tool_list: Annotated[list[type[BaseTool]], Depends(get_tool_list)],
```
jankrepl, line 109: As discussed, I would make sure that all our endpoints are protected

frontend/src/components/chat-input.tsx
```diff
@@ -54,26 +73,36 @@ export function ChatInput() {
...
         className="w-full flex justify-center"
       >
-        <div className="relative w-1/2">
+        <div className="flex items-center min-w-[70%] max-w-[100%] border-2 border-gray-500 rounded-full overflow-hidden">
           <input
             name="content"
             type="text"
             autoComplete="off"
-            className="border-2 border-gray-500 w-full p-4 rounded-full"
+            className="flex-grow p-4 outline-none bg-transparent text-white"
```
jankrepl, line 81: Careful, in the light mode the input text is not visible

Screenshot 2025-02-07 at 10 57 10


I guess you can use to have a separate value for light/dark. Or some other trick:)

frontend/src/components/chat-page.tsx
```diff
@@ -192,10 +212,10 @@ export function ChatPage({ threadId, initialMessages }: ChatPageProps) {
         className="flex flex-col justify-center items-center gap-4 mb-4"
         onSubmit={handleSubmit}
       >
-        <div className="relative w-1/2">
+        <div className="flex items-center min-w-[70%] max-w-[100%] border-2 border-gray-500 rounded-full overflow-hidden">
```
jankrepl, line 215: I don't like the fact that if one types long text it is behind the buttons


Screenshot 2025-02-07 at 11 05 44

Note that `chat-input` does not have this behavior

frontend/src/components/chat-input.tsx
WonderPG: Nice catch, indeed the `text-white` shouldn't be here. Removing it makes it compatible with both. Done in f607e35

backend/src/neuroagent/app/routers/tools.py
WonderPG: Done in f607e35. I didn't add the `thread_id` here since to me this is highly unrelated to a thread. However I added 
```python
_: Annotated[str, Depends(get_user_id)]
```
which checks the bearer token at least. Is it okay ?

frontend/src/components/chat-page.tsx
WonderPG: Fixed it this morning, you can pull 😄

WonderPG: I believe it was in 2a1324d

backend/src/neuroagent/app/routers/tools.py
jankrepl: sure:) thanks

jankrepl:
Overall looks great:) I am fine with merging it and iterating on it in the future.

IMO the next follow up PR could be about actually listing the tools and providing their description - https://github.com/openbraininstitute/neuroagent/issues/44. The current tool titles (which are actually slugs) only have meaning to us and a user will very likely not know what they are (unless they ask the LLM)

Thank you @WonderPG

jankrepl (approved):

BoBer78 (approved):
Thank you for cleaning the divs and removing the absolute and translates ! 
LGTM !

PR #94, opened 2025-02-07
Improve hil refusal

### LINKED ISSUE #46: Prompt engineer HIL rejected tool result
Right now it says `The tool call has been invalidated by the user.` which might not discourage the LLM from re-requesting the same tool call in the next round

WonderPG:
Closes #46


jankrepl:
The feature makes a lot of sense. Personally, I am not super sure about having a dialog on top of another dialog.

I think a more appropriate component would be a "stepper". ShadCN does not have it so it would require some customization /copy pasting.

WonderPG:
Boris had a similar comment and I do agree that a dialog on top of another dialog is not the best. However I was not too sure about what replacement there could be. I will investigate a bit

BoBer78 (dismissed):
🦫

BoBer78 (approved):

PR #112, opened 2025-02-11
Better settings on first login

### LINKED ISSUE #98: Virtual lab + Project missing in settings -> no indicator
First time users might be confused because they haven't populated the project/virtual lab in `settings` and there is no indication that they need to do it

### LINKED ISSUE #101: Redirect to `/` after saving `/settings`

BoBer78:
closes #101  closes #98

### DIFF
--- a/frontend/src/app/page.tsx
+++ b/frontend/src/app/page.tsx
@@ -1,6 +1,18 @@
 import { ChatInput } from "@/components/chat-input";
 import { fetcher } from "@/lib/fetcher";
 import { auth } from "@/lib/auth";
+import { cookies } from "next/headers";
+import { redirect } from "next/navigation";
+
+async function isVlabAndProjectPopulated() {
+  const cookieStore = await cookies();
+  const vlabId = cookieStore.get("virtualLabID");
+  const projectId = cookieStore.get("projectID");
+
+  if (!(vlabId && projectId)) {
+    redirect("/settings");
+  }
+}
 
 async function getToolList() {
   const session = await auth();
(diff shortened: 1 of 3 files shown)


frontend/src/app/page.tsx
```diff
@@ -1,6 +1,18 @@
...
 import { fetcher } from "@/lib/fetcher";
 import { auth } from "@/lib/auth";
+import { cookies } from "next/headers";
+import { redirect } from "next/navigation";
+
+async function isVlabAndProjectPopulated() {
+  const cookieStore = await cookies();
+  const vlabId = cookieStore.get("virtualLabID");
+  const projectId = cookieStore.get("projectID");
+
```
jankrepl, line 11: Check `src/lib/cookies-server.ts`

frontend/src/app/page.tsx
```diff
@@ -1,6 +1,18 @@
...
+  if (!(vlabId && projectId)) {
+    redirect("/settings");
```
jankrepl, line 13: As mentioned, `redirect` is a bit buggy in docker compose...but it should be fine

frontend/src/app/page.tsx
BoBer78: Fixed in [733948f](https://github.com/openbraininstitute/neuroagent/pull/112/commits/733948f1f1d174d3dbd028199fdea3295a1d4778).

frontend/src/app/page.tsx
BoBer78: I tested it, seems to work properly.

jankrepl: yeh...but we can do that later:)

jankrepl (approved):
Looks great:) Let's merge and we can add some more complicated logic in a future PR

jankrepl (approved):

PR #117, opened 2025-02-13
Allow user scrolling when streaming.

### LINKED ISSUE #71: Allow user to scroll up when LLM is streaming
Currently if the LLM is streaming its answer, the UI will always scroll back to the last generated token. To read anything you have to wait for the LLM to finish then scroll up

BoBer78:
closes #71


WonderPG (approved):
Much better experience, love it

PR #118, opened 2025-02-13
Return json

### LINKED ISSUE #63: Tool results should be valid JSONs (or arbitrary strings)
(See update below) - TLDR I think this specific tool does not dump a valid JSON
Some tools have valid JSON as return value.

~In the frontend, it would be nice to print it out nicely in that case (basically what Postman calls beautify). Currently, we just put it on a single line which makes it basically unreadable~

Image

Note that the problem is very likely there because of the single quotes

~I think the current code does try to somehow achieve it but it clearly fails (unless I am missing something and we don't return a valid JSON for that tool)~

https://github.com/openbraininstitute/neuroagent/blob/81779c339fba77007b63aec31241cca3dae0c61f/frontend/src/components/chat-message-tool.tsx#L215-L226

[UPDATE 1] Actually, I try the weather tool and the result there is a valid JSON and the frontend beautifies it without issues. So I am assuming this is a backend problem?

Image

WonderPG:
Closes #63

### DIFF
--- a/backend/src/neuroagent/agent_routine.py
+++ b/backend/src/neuroagent/agent_routine.py
@@ -75,9 +75,13 @@ class AgentsRoutine:
                 )
             case _:
                 try:
+                    return Result(
+                        value=result if isinstance(result, str) else json.dumps(result)
+                    )
+                except json.JSONDecodeError:
                     return Result(value=str(result))
                 except Exception as e:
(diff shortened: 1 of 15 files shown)


jankrepl:
Tried locally and works great:) feel free to merge `main` and then I will approve. @WonderPG

jankrepl (approved):

PR #122, opened 2025-02-13
Disable input when HIL

### LINKED ISSUE #116: Block the chat input when the user is not supposed to enter stuff
Currently it is possible to write and send messages when the LLM is writing, but also when there is a HIL validation pending. If the user does so, it completely crashes the conversation.

In my opinion the user should be allowed to write in these moments, but not to send. It would be good to deactivate the submission of the form as soon as the user is not supposed to send it anymore

BoBer78:
closes #116


WonderPG (commented):
Works great, but do you think it would be a lot of extra logic to allow to type, but not to send while the LLM talks/HIL required ? If yes we can address that in another PR

jankrepl (approved):

PR #123, opened 2025-02-13
Push to ecr prod+staging

WonderPG:


.github/workflows/ecr.yml
```diff
@@ -10,6 +10,9 @@ permissions:
   contents: read    # This is required for actions/checkout
 
 jobs:
+  defaults:
+    run:
+      working-directory: backend
```
WonderPG, line 15: This can change in the future if we decide to push the frontend as well. To avoid unecessarily large build context I decided to cd inside of the folder rather than pointing to the backend image from root.

jankrepl: Sounds good:)

jankrepl (approved):

BoBer78 (approved):

BoBer78 (approved):

PR #130, opened 2025-02-14
Tools + content fix

### LINKED ISSUE #89: AI tool with content
We discussed it many times. There is a PR that partially addresses it.

I just want to create a ticket for it since it breaks the current application when you get both. What is tricky is that it is hard to reproduce it and therefore I cannot add a screen recording.

#2

BoBer78:
closes #89


jankrepl (dismissed):
Approving (and can reapprove once needed).

Assuming this really solves all the problems we need it even more since with the plotting it is very common to have both content and tool call

Thanks @BoBer78

jankrepl (approved):

PR #132, opened 2025-02-17
Better first messsage logic

### LINKED ISSUE #99: Better first message logic
I feel like our logic is kinda flawed right now : 

- we are on `chat-input.tsx` 
- when triggering the chat we create a new thread and set the `newMesage` in the store
- once the thread has been created, we redirect
- on redirect, we land on `threads/[ThreadId]/page.tsx`
- this fetches the messages from the DB, and the tool list. **This is useless and takes time**
- We arrive in the `chat-page.tsx`, add the `newMesage` from store , initialise the `useChat` hook.
- Finally we trigger the chat. 

I would add a useChat hook in the `chat-input.tsx` (or in the parent, and share it). 
Chat input already has the new message, we can wait for the thread to be created and then instantly trigger the chat. 
Once the streaming is finished we redirect to the new page. 

Then the issue becomes that when we redirect to the new thread we need to re-feth from the DB ... 

What do you guys think ?

BoBer78:
closes #99


BoBer78:
This is a first POC that would improve the logic for the first message rendering. I'm not sure the added complexity is worth the improved loading time. I'll try to polish in and then we can decide.

PR #133, opened 2025-02-17
Display the entire json when calling tools

### LINKED ISSUE #105: Make sure all of the tool inputs are displayed, not only LLM generated ones
Sometimes the LLM doesn't spit every parameter (which is expected) and only specifies the ones that are not default. However the frontend shows only the LLM generated parameters and not all of them

Image

It is a bit misleading for the user especially in the case of HIL, where he would have to know the name of the missing key to change it.

We solved this issue in the past by using pydantic in the backend, e.g.
```python
entire_json = ToolClassInput(**kwargs).model_dump_json()
```
would return the full json. `**kwargs` would contain only the LLM generated inputs.

WonderPG:
Closes #105


backend/src/neuroagent/agent_routine.py
```diff
@@ -223,10 +224,13 @@ async def astream(
 
                     elif choice.finish_reason == "tool_calls":
                         for tool_call in draft_tool_calls:
+                            input_schema = tool_map[tool_call["name"]].__annotations__[
```
jankrepl, line 227: edge case comment

If the LLM messes up and gives wrong parameters...there will be an uncaught error. Or I mean the error will not go to tool result.

WonderPG: I did a quick and dirty fix in 945dfaa where I try except the pydantic class instantiation and in case of validation error default to outputting what the LLM provided. Not very clean but it should very rarely happen.

BoBer78 (approved):

PR #143, opened 2025-02-20
Better loading and first message

### LINKED ISSUE #99: Better first message logic

BoBer78:
closes #99


jankrepl (approved):

frontend/src/components/sidebar/sidebar.tsx
```diff
@@ -21,6 +21,7 @@ export function Sidebar({ children }: { children: React.ReactNode }) {
   if (pathname === "/login") {
     return null;
   }
+  console.log("Re-render sidebar");
```
jankrepl, line 24: Let's remove this one.

frontend/src/components/sidebar/thread-list.tsx
```diff
@@ -33,7 +32,7 @@ async function getThreads(): Promise<BThread[]> {
 
 export async function ThreadList() {
   const threads = await getThreads();
-
+  console.log(threads);
```
jankrepl, line 35: again, let's remove this one (and all the other `console.log` if there are more)

jankrepl (approved):
LGTM:) Thank you:)

jankrepl (approved):

PR #155, opened 2025-02-25
Fix bug in docker compose and add back redirects

BoBer78:


BoBer78:
Seems like the solution is documented in a couple places : 
https://stackoverflow.com/questions/75841179/nextjs-api-axios-nextauth-issue-axioserror-connect-econnrefused-13000/75842143#75842143
Very strange ...

PR #163, opened 2025-02-28
Ollama support (optional - not the new default)

### LINKED ISSUE #83: Ollama inside docker compose (POC)
Not saying we have to merge this and I am not saying we should just start this crazy side project of finding the best local LLM. 

However, Ollama https://ollama.com/blog/openai-compatibility and many other tools/models that are opensource should mirror the `chat/completions` API and therefore  one can use the openai SDK out of the box.  We just need to change the `base_url` in the `AsyncClient` ( https://github.com/openai/openai-python?tab=readme-ov-file#configuring-the-http-client)

Why useful? For local development.

jankrepl:
Closes #83 

# For reviewers
* All instructions are in the `README.md`
* I set things up so that Ollama is optional and by default we continue with the current OpenAI setup - I introduced a docker compose profile called `localllm` that will optionally spin up Ollama (more info https://event-driven.io/en/docker_compose_profiles/) 
*  There is a problem: As soon as we provide 1+ tools streaming stops working (even if we stream just a simple `ai-message` without tools) See bottom of the page for more if you are interested
  
 

# Tools + Streaming details
AFAIK this is a known "issue" that Ollama has and will be fixed in the future. I actually thought it would not be as extreme - I thought that the `ai-message`  without tool calls would be streamed and the `ai-tool` would not be streamed but we would receive the full call once we have it. I was wrong (unless I messed something up)

**References**
* https://github.com/ollama/ollama/issues/7886
* https://github.com/ollama/ollama/issues/5796#issuecomment-2508776772

**Reproducible example of the issue**
```python
import asyncio
from openai import AsyncOpenAI

from neuroagent.tools import WeatherTool

async def main():
    client = AsyncOpenAI(base_url="http://localhost:11434/v1", api_key="fake_token")

    tools = [WeatherTool.pydantic_to_openai_schema()]

    stream = await client.chat.completions.create(
        model="qwen2.5:0.5b",
        messages=[{"role": "user", "content": "Can you tell me a longer joke"}],
        stream=True,
        # tools=tools,  # if you comment out everything is streamed, if you keep it there just 1 chunk
    )
    async for chunk in stream:
        for choice in chunk.choices:
            if choice.delta.content is not None:
                print(choice.delta.content, end="")
                print("---------------NEW CHUNK---------------")

if __name__ == "__main__":
    asyncio.run(main())


jankrepl:
Closing since this is not the "business" direction we are taking

PR #185, opened 2025-03-06
Single agentic step endpoint

### LINKED ISSUE #183: Create a "one-shot" endpoint
Currently the frontend doesn't need to persist the threads. We should create a one shot endpoint that accepts an entire history and returns the new message

WonderPG:
Closes #183 

To test it, replace the `useChat` hook in `chat-page.tsx` with the following
```javascript
const {
    messages: messagesRaw,
    input,
    handleInputChange,
    handleSubmit,
    error,
    isLoading,
    setMessages: setMessagesRaw,
  } = useChat({
    api: `${env.NEXT_PUBLIC_BACKEND_URL}/qa/streamed_completion`,
    headers: {
      Authorization: `Bearer ${session?.accessToken}`,
    },
    initialMessages,
    experimental_throttle: 100,
    experimental_prepareRequestBody: ({ messages }) => {
      const selectedTools = Object.keys(checkedTools).filter(
        (key) => key !== "allchecked" && checkedTools[key] === true,
      );
      return { messages: messages, tool_selection: selectedTools };
    },
  });
```
With the current frontend implementation, it will still create a thread but won't store the messages. If you refresh the conversation still exists but will be empty. You can continue sending messages in this same conversation. However it does that because we decided to fire the thread creation when sending the first message. In the real frontend they will just not create the thread.

Disclaimer: Every tool related to `thread_id`, `vlab_id` or `project_id` will crash. They are not meant to be used with this endpoint.

### DIFF
--- a/backend/src/neuroagent/agent_routine.py
+++ b/backend/src/neuroagent/agent_routine.py
@@ -180,17 +180,17 @@ class AgentsRoutine:
     async def astream(
         self,
         agent: Agent,
-        messages: list[Messages],
+        messages: list[Messages] | list[dict[str, Any]],
         context_variables: dict[str, Any] = {},
         model_override: str | None = None,
         max_turns: int | float = float("inf"),
     ) -> AsyncIterator[str]:
         """Stream the agent response."""
         active_agent = agent
-        content = (
+        content: list[dict[str, Any]] = (
             messages
             if isinstance(messages[0], dict)
-            else await messages_to_openai_content(messages)
+            else await messages_to_openai_content(messages)  # type: ignore
         )
         history = copy.deepcopy(content)
         init_len = len(messages)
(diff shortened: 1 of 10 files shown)


backend/src/neuroagent/agent_routine.py
```diff
@@ -312,29 +316,31 @@ async def astream(
...
                     order=len(messages),
-                    thread_id=messages[-1].thread_id,
+                    thread_id="temp"
+                    if isinstance(messages[-1], dict)
+                    else messages[-1].thread_id,
                     entity=get_entity(message),
                     content=json.dumps(message),
                     tool_calls=tool_calls,
                 )
             )
```
WonderPG, line 328: The ugly hack is here, the SQL class is getting appended to the `messages` list, which can be a list of openai's compatible dictionaries. The rest of the code is assuming `messages` contains SQL objects, but it uses `messages[-1]` everywhere so it works. Objectively we have a mix of dictionaries and SQL objects in the list, but we just yield the vercel JSONs and discard the `messages` list.

WonderPG:
Closing due to deprecation since #187

PR #187, opened 2025-03-06
Enable threads outside of vlabs

### LINKED ISSUE #183: Create a "one-shot" endpoint

WonderPG:
Closes #183 but in a funny way


jankrepl (dismissed):
LGTM:)

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Object storage + plotting logic.
 - Keywords filtering for literature search.
 - [frontend] - bugfixes.
+- Single agentic step endpoint.
```
BoBer78, line 62: I guess this is still from the old PR.

backend/alembic/env.py
```diff
@@ -79,6 +79,7 @@ def run_migrations_offline() -> None:
         target_metadata=target_metadata,
         literal_binds=True,
         dialect_opts={"paramstyle": "named"},
+        render_as_batch=True,
```
BoBer78, line 82: Small question, what is it for ?

WonderPG: SQLite doesn't allow for `ALTER` statements which modify an existing table. This argument tells alembic to instead create a temp tab, migrate the data, remove the old tab then delete it.

CHANGELOG.md
WonderPG: Changed in 3f45218

jankrepl (approved):

PR #195, opened 2025-03-10
Bugfixes

### LINKED ISSUE #156: Markdown tables still break the UI sometimes.
If you ask this question : 

Image

And then ask for a table, it will sometimes make it horizontal and very big : 

Image

This is due to that : https://github.com/openbraininstitute/neuroagent/blob/73239b71224547a54296ca223f920903e5020301/frontend/src/components/chat/chat-message-ai.tsx#L47 
Changing this to a `<div>` fixes it. Then some more tweaking to the rest of the UI needs to be done.

### LINKED ISSUE #157: Shift + Enter should add new lines to the input
In #149 we made sure `Shift + Enter` does not submit.

However, I suggest we try to copy ChatGPT (see the screenrecording) and make sure that it not only does not submit but also it adds a new line to the input.

Note that the behavior should be the same in the `/` and in `/threads/{threadId}`

https://github.com/user-attachments/assets/ad6f026a-e642-44f6-a301-22b7858d0e20

### LINKED ISSUE #164: Use structured output to generate the title
So I used a different LLM (llama3.2) and the title generation endpoint contained the suggested title but also some AI comments

> It seems like you're starting our conversation! Here's a possible short title:"Hello from AI Companion"

I suggest we use structured output to guarantee we only get what we need

https://platform.openai.com/docs/guides/structured-outputs

https://github.com/openbraininstitute/neuroagent/blob/a05a263a5b9c8b1316c66104d60322a98eda629b/backend/src/neuroagent/app/routers/threads.py#L86-L96

### LINKED ISSUE #179: Error message in production are not shown properly.
Image

In `app/threads/[threadId]/error.tsx` i have made it so that it displays the error message. It works in `dev` but not in docker compose. 
We can add a generic error message and then add `notFound.tsx` for 404 errors (https://nextjs.org/learn/dashboard-app/error-handling).

### LINKED ISSUE #191: Editing title of recent thread proposes wrong text
https://github.com/user-attachments/assets/3ca906b6-ef27-4eae-a673-f82e92de4f82

The `New Chat` should not be there.

As soon as one refreshes (and refetches the sidebar) it shows the correct title

### LINKED ISSUE #194: Specify species in the hardcoded default questions
https://github.com/openbraininstitute/neuroagent/blob/0c7493a8af2c02d89bc21523f2b2e184bd526339/frontend/src/components/chat/question-suggestion-cards.tsx#L10-L24

We should just add "of rodents" at the end of all of those. Note that there will probably be an independent ticket for adding the species to the user journey (if not you can directly do it here)

BoBer78:
closes #194  closes #156 closes #157 closes #179  closes #191  closes #164

### DIFF
--- a/backend/tests/mock_client.py
+++ b/backend/tests/mock_client.py
@@ -1,13 +1,7 @@
 import json
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, Mock
 
-from openai.types.chat import (
-    ChatCompletionMessage,
-    ParsedChatCompletion,
-    ParsedChatCompletionMessage,
-    ParsedChoice,
-)
-from openai.types.chat.chat_completion import ChatCompletion, Choice
+from openai.types.chat.chat_completion import ChatCompletion
 from openai.types.chat.chat_completion_message_tool_call import (
     ChatCompletionMessageToolCall,
     Function,
@@ -18,7 +12,6 @@ def create_mock_response(
     message,
     function_calls=[],
     model="gpt-4o-mini",
-    is_structured_output=False,
     structured_output_class=None,
 ):
     role = message.get("role", "assistant")
@@ -38,82 +31,52 @@ def create_mock_response(
         if function_calls
         else None
     )
-    if not is_structured_output:
-        return ChatCompletion(
-            id="mock_cc_id",
-            created=1234567890,
-            model=model,
-            object="chat.completion",
-            choices=[
-                Choice(
-                    message=ChatCompletionMessage(
-                        role=role, content=content, tool_calls=tool_calls
-                    ),
-                    finish_reason="stop",
-                    index=0,
-                )
-            ],
-        )
-    else:
-        return ParsedChatCompletion(
-            id="mock_cc_id",
-            created=1234567890,
-            model=model,
-            object="chat.completion",
-            choices=[
-                ParsedChoice(
-                    finish_reason="stop",
-                    index=0,
-                    logprobs=None,
-                    message=ParsedChatCompletionMessage(
-                        content=json.dumps({"title": content}),
-                        refusal=None,
-                        role=role,
-                        audio=None,
-                        function_call=None,
-                        tool_calls=tool_calls,
-                        parsed=structured_output_class,
-                    ),
-                )
-            ],
-        )
+    return Mock(
+        id="mock_cc_id",
+        created=1234567890,
+        model=model,
+        object="chat.completion",
+        choices=[
+            Mock(
+                message=Mock(
+                    role=role,
+                    content=content,
+                    tool_calls=tool_calls,
+                    parsed=structured_output_class,
+                ),
+                finish_reason="stop",
+                index=0,
+            )
+        ],
+    )
 
 
 class MockOpenAIClient:
-    def __init__(self, is_structured_output=False):
-        # If this is set to true, the mock acts for structured output.
-        self.is_structured_output = is_structured_output
-
-        if not is_structured_output:
-            self.chat = AsyncMock()
-            self.chat.completions = AsyncMock()
-        else:
-            self.beta = AsyncMock()
-            self.beta.chat = AsyncMock()
-            self.beta.chat.completions = AsyncMock()
+    def __init__(self):
+        self.chat = AsyncMock()
+        self.chat.completions = AsyncMock()
+        self.beta = AsyncMock()
+        self.beta.chat = AsyncMock()
+        self.beta.chat.completions = AsyncMock()
 
     def set_response(self, response: ChatCompletion):
         """
         Set the mock to return a specific response.
         :param response: A ChatCompletion response to return.
         """
-        if not self.is_structured_output:
-            self.chat.completions.create.return_value = response
-        else:
-            self.beta.chat.completions.parse.return_value = response
+        self.chat.completions.create.return_value = response
+        self.beta.chat.completions.parse.return_value = response
 
     def set_sequential_responses(self, responses: list[ChatCompletion]):
         """
         Set the mock to return different responses sequentially.
         :param responses: A list of ChatCompletion responses to return in order.
         """
-        if not self.is_structured_output:
-            self.chat.completions.create.side_effect = responses
-        else:
-            self.beta.chat.completions.parse.side_effect = responses
+        self.chat.completions.create.side_effect = responses
+        self.beta.chat.completions.parse.side_effect = responses
 
     def assert_create_called_with(self, **kwargs):
-        if not self.is_structured_output:
-            self.chat.completions.create.assert_called_with(**kwargs)
-        else:
-            self.beta.chat.completions.parse.assert_called_with(**kwargs)
+        self.chat.completions.create.assert_called_with(**kwargs)
+
+    def assert_create_called_with_structure_output(self, **kwargs):
+        self.beta.chat.completions.parse.assert_called_with(**kwargs)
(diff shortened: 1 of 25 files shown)


frontend/src/lib/fetcher.ts
```diff
@@ -64,6 +65,9 @@ export async function fetcher({
   });
 
   if (!response.ok) {
+    if (response.status === 404) {
+      return notFound();
```
jankrepl, line 69: I don't like the idea of calling the `notFound()` here. 

Looking at the official docs https://nextjs.org/docs/app/api-reference/functions/not-found they clearly put this inside the `page.tsx` - or in other words it is done on the "Page" level. This `fetcher.ts` was designed to implement data fetching logic completely independently of react/nextjs. One example of that is that  `frontend/src/hooks/fetch.ts` wraps the fetcher inside react query. And there raising the exception is key to get correct status information and retries.

Alternative (chatgpt assisted)

Define a new subclass of `Error` and then include the statusCode inside and then you can always try/catch when you call it

```javascript
class CustomError extends Error {
    constructor(message, statusCode) {
        super(message); // Call the parent constructor
        this.statusCode = statusCode; // Add custom property
        this.name = this.constructor.name; // Set error name dynamically
        Error.captureStackTrace(this, this.constructor); // Preserve stack trace
    }
}

// Example usage:
try {
    throw new CustomError("Not Found", 404);
} catch (error) {
    console.log(error.message);     // "Not Found"
    console.log(error.statusCode);  // 404
    console.log(error.stack);       // Full stack trace
}

```

BoBer78: I personally hate the fact that the fetcher implements the error handling. I tried before to make this custom error class and NextJS transforms it back to a normal error and I lost all the additional parameters.
I'll try to put the `NotFound` in page.tsx, and not the fetcher.

jankrepl: Point taken @BoBer78 

Alternative 2) return the error and have two return types (one for success and the other one for error)

Alternative 3) don't use this fetcher at all..just write a raw `fetch` whenever you data. Not pointing figures but @WonderPG pushed for this approach of having some "fetching" wrapper that every component is forced to use but IMO it assumes that your wrapper fits every single use case which imo is hard

BoBer78: I'll try your solution 1), and then if it is not to complicated we can go with it !

backend/tests/mock_client.py
```diff
@@ -27,42 +38,82 @@ def create_mock_response(message, function_calls=[], model="gpt-4o-mini"):
...
+                        refusal=None,
+                        role=role,
+                        audio=None,
+                        function_call=None,
+                        tool_calls=tool_calls,
+                        parsed=structured_output_class,
+                    ),
+                )
+            ],
+        )
```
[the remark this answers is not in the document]
BoBer78: I like the idea, I'll try to implement it like that !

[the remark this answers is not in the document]
BoBer78: Done in [981aebb](https://github.com/openbraininstitute/neuroagent/pull/195/commits/981aebbd29dfbe4f572afc789c271b0dca31f827).

frontend/src/lib/fetcher.ts
BoBer78: Done in [dbbabb1](https://github.com/openbraininstitute/neuroagent/pull/195/commits/dbbabb11ddbff01669a9e1a3098e4c48c60dec9e). I used solution 1 and I put notFond in the pages.

jankrepl:
Regarding the multiline it is almost perfect - can you please just make sure there is enough padding? The text should not "touch" the border


https://github.com/user-attachments/assets/b5be13f2-165a-416a-8764-85127f516d40

frontend/src/app/tools/[toolName]/not-found.tsx
```diff
@@ -0,0 +1,37 @@
...
+  reset,
+}: {
+  error: Error & { digest?: string };
+  reset: () => void;
+}) {
+  useEffect(() => {
+    // Log the error to an error reporting service
+    console.error(error);
+  }, [error]);
+
```
Comment (jankrepl) [inline frontend/src/app/tools/[toolName]/not-found.tsx:18]: So we need to keep this one in here? I would remove it everywhere.


https://github.com/user-attachments/assets/b451be3e-1329-4431-8650-76309ce47e86

jankrepl:
Amazing job:) I tried it locally and left some minor comments above. I will approve once you address them

@BoBer78

jankrepl:
One thing that surprised me was that we do allow for multiline input but then somehow the new lines are lost inside the `chat-message-human`...is that on purpose?

ChatGPT does preserve them


https://github.com/user-attachments/assets/9a4d61ad-8cfa-4b38-af6e-d26e06b359b0

BoBer78:
> One thing that surprised me was that we do allow for multiline input but then somehow the new lines are lost inside the `chat-message-human`...is that on purpose?
> 
> ChatGPT does preserve them
> Screen.Recording.2025-03-12.at.12.35.07.mov

My guess is that we should also support some kind of formating in the Humain message box. Maybe even markdown

jankrepl:
> > One thing that surprised me was that we do allow for multiline input but then somehow the new lines are lost inside the `chat-message-human`...is that on purpose?
> > ChatGPT does preserve them
> > Screen.Recording.2025-03-12.at.12.35.07.mov
> 
> My guess is that we should also support some kind of formating in the Humain message box. Maybe even markdown

I don't think ChatGPT supports markdown for human messages. Might be an overkill. But I would definitely support new line.

BoBer78:
> > > One thing that surprised me was that we do allow for multiline input but then somehow the new lines are lost inside the `chat-message-human`...is that on purpose?
> > > ChatGPT does preserve them
> > > Screen.Recording.2025-03-12.at.12.35.07.mov
> > 
> > 
> > My guess is that we should also support some kind of formating in the Humain message box. Maybe even markdown
> 
> I don't think ChatGPT supports markdown for human messages (at least I did not manage to trigger it). Might be an overkill. But I would definitely support new line.

I added line breaks in the human message, it was a simple Tailwind class

frontend/src/app/threads/[threadId]/not-found.tsx
```diff
@@ -0,0 +1,23 @@
+"use client";
+
+import { Button } from "@/components/ui/button";
+import Link from "next/link";
+
+export default function NotFound() {
+  return (
+    <div className="container mx-auto px-4 py-6 h-[calc(100vh-4rem)] flex flex-col items-center justify-center">
+      <div className="text-center space-y-4">
+        <h2 className="text-2xl font-bold">404 not found.</h2>
```
Comment (jankrepl) [inline frontend/src/app/threads/[threadId]/not-found.tsx:10]: Detail.

I would just unify this with `tools/[toolName]/not-found.tsx`


There we say `404 Not Found` (here we say `404 not found.`)

frontend/src/app/tools/[toolName]/not-found.tsx
```diff
@@ -0,0 +1,25 @@
...
+        <p className="text-muted-foreground">
+          The requested tool could not be found.
+        </p>
+        <div className="flex gap-4 justify-center mt-6">
+          <Link href="/tools">
+            <Button>Return to Tools</Button>
+          </Link>
+          <Button variant="outline" onClick={() => reset()}>
+            Try again
+          </Button>
```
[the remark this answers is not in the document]
Comment (BoBer78) [reply frontend/src/app/tools/[toolName]/not-found.tsx:20]: Should I also remove it in the `error.tsx` page ?

jankrepl (approved):

PR #196, opened 2025-03-10
Multiagent poc

### LINKED ISSUE #165: Multiagent POC
Let's introduce multiple agents. It would be nice to directly adjust the frontend

Some ideas
* color tool calls based on agent
* Group `tools` based on the agent

WonderPG:
Closes #165 

The global behavior of the multi agent relies quite a bit on the system prompts provided to each agent. As you can imagine I tried to do my best but it might still need some tweaking. Please try as many workflows as possible and check if it executes it correctly. If you encounter issues, feel free to adjust the system prompt.

You can control the multi/simple behavior through this env var
```python
NEUROAGENT_AGENT__COMPOSITION=
```
set it to "multi" or "simple" to switch between these.


backend/src/neuroagent/app/config.py
```diff
@@ -4,14 +4,30 @@
...
-from pydantic import BaseModel, ConfigDict, SecretStr
+from pydantic import BaseModel, ConfigDict, SecretStr, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
 class SettingsAgent(BaseModel):
     """Agent setting."""
 
-    model: Literal["simple", "multi"] = "simple"
+    composition: Literal["simple", "multi"] = "multi"
```
jankrepl, line 14: Discussed already but this is probably the biggest "red flag" I am seeing in here. From the conceptual point of view. I am almost sure that if we merge this then we will just end up maintaining one of the two.

PR #200, opened 2025-03-11
Comstic Fixes

### LINKED ISSUE #97: Spinner after clicking sign in
Currently, after you click "Sign in" the UI does not convey any pending progress. I suggest we add a spinner 

https://github.com/user-attachments/assets/2bb3d4ab-8bef-45af-9067-ac6d0b68d612

### LINKED ISSUE #166: Make the validation button more explicit
Feedback from Michael: It is not obvious what to do once a tool requires validation.

We should make it more explicit to the user where to click to perform the validation

### LINKED ISSUE #192: Tooltip on thread hover (showing full title)
Right now there is no way to show the full title other than clicking edit. And getting a title that ends up being truncated is quite common. I suggest we implement a tooltip on hover. However, make sure to set the   `delayDuration` to something reasoanble (something bigger than just a quick hover when selecting a thread but something smaller than eternity)

Use https://ui.shadcn.com/docs/components/tooltip which uses https://www.radix-ui.com/primitives/docs/components/tooltip#api-reference

BoBer78:
closes #97 closes #166  closes #192

### DIFF
--- a/frontend/src/components/sidebar/thread-card-sidebar.tsx
+++ b/frontend/src/components/sidebar/thread-card-sidebar.tsx
@@ -11,12 +11,6 @@ import { DeleteThreadDialog } from "@/components/sidebar/delete-thread-dialog";
 import { Button } from "@/components/ui/button";
 import { useState, useOptimistic } from "react";
 import {
-  Tooltip,
-  TooltipContent,
-  TooltipProvider,
-  TooltipTrigger,
-} from "@/components/ui/tooltip";
-import {
   DropdownMenu,
   DropdownMenuTrigger,
   DropdownMenuContent,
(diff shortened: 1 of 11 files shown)


BoBer78:
A couple of comments for the reviewers. 
- I added a simple title to the div of the side bar. This allows to show the full title while hovering it. If we want more control we need to make the code more complex or add an external library to tweak the tooltip UI.
- The way the current authentification works is still a mystery to me ... I have done a simple fix to show a spinner that would in principle be enough for 99% of usecases. Happy to discuss this one, I'm not happy with the solution.

frontend/src/components/sidebar/thread-card-sidebar.tsx
```diff
@@ -44,6 +44,7 @@ export function ThreadCardSidebar({ title, threadId }: ThreadCardSidebarProps) {
           "bg-accent": currentThreadId === threadId,
         },
       )}
+      title={title}
```
jankrepl, line 47: haha...did not know this trick:) This seems to be a poor man's way how to show something on hover.


Interesting. I mean it does the job - it is not as pretty as the shadcn tooltip but I am fine with this solution. Your call

BoBer78: I did not want to add much more complexity for such a small issue. I'll try to find a simple solution and if it start to take too much time I'll keep this simple fix.

jankrepl: sounds good:) Resolving.

jankrepl (approved):
LGTM:) Thank you a lot !!! @BoBer78

PR #206, opened 2025-03-14
Make s3 compatible

jankrepl:
The current `main` was written in a way that by default we use MinIO (very convenient for local development). However, now that we want to deploy on S3 I am changing the config so that by default we go to S3 (can be overridden via env var) and we use credentials provided outside of Python.

So locally to use minio, make sure 

```
NEUROAGENT_STORAGE__ENDPOINT_URL=http://localhost:9000
NEUROAGENT_STORAGE__ACCESS_KEY=minioadmin
NEUROAGENT_STORAGE__SECRET_KEY=minioadmin
```

Make sure the configuration makes it possible to swap out s3/minIO


WonderPG (approved):
LGTM

Issue #211, opened 2025-03-17
Whitelist tools via env var


jankrepl:
```
NEUROAGENT_TOOLS_WHITELIST=literature-search-tool,web-search-tool
```

if not provided then we allow all tools

if we go for multiagent maybe there is a better mechanism (starting agent)

PR #215, opened 2025-03-17
Allow blacklisting certain tools and add Now tool

### LINKED ISSUE #214: Include get now tool
Up until now we have been using it just for debugging. However, I would argue we should include it in our deployment since we definitely expect the user to include things like "recent", "last year", "last month", ... in their questions. Some of our tools (e.g. literature search) then expect the date in inputs. The LLM will always point to an older date than "today" since it has a knowledge cutoff.

WonderPG:
Closes #214 #211 

I went with a blacklist approach instead, We can leave it `None` if nothing is blacklisted and no need to always append the expanding list of validated tools.


jankrepl (dismissed):
LGTM:)

BoBer78 (approved):
LGTM !

PR #217, opened 2025-03-18
Display plots inside thread

### LINKED ISSUE #160: Plot thumbnail inside `/threads/{threadId}`
Currently, the only way to view the plots is via `/viewer/{storageId}`. However, it would be nice to show some "smaller"/thumbnail version of the plot directly in the chat.

You can approach this in any way you want. One low hanging way to do this is to just introduce something like `chat-message-plot.tsx` and insert this component whenever the previous message was `ai-tool` where the tool slug was `plot-generator`. Edge case is that there could have been multiple parallel tool calls so you need to generate the `chat-message-plot.tsx` for each of them. If you manage to reuse the component in `src/components/plots` (feel free to refactor) and just make a smaller version of it that would be great.

### LINKED ISSUE #219: Line chart x limit is always (0, something)
No matter what the datapoints actually are the x axis will have a left bound of 0. IMO it should be dynamic and probably equal to something closer to `min(x for x,_ in data)`. But maybe charts.js can just imply it automatically. Please go for the best solution.

Image

BoBer78:
closes #219  closes  #160


backend/src/neuroagent/app/dependencies.py
```diff
@@ -206,19 +206,12 @@ def get_starting_agent(
...
-        if settings.misc.frontend_url
-        else "Never try to generate links to internal storage ids"
-    )
+                Do no blindly repeat the brain region requested by the user, use the output of the tools instead.
+                Never try to generate links to internal storage ids"""
 
     agent = Agent(
         name="Agent",
-        instructions=f"{base_instructions}\n{storage_instructions}",
+        instructions=f"{base_instructions}",
```
WonderPG, line 214: No need for the f string

frontend/src/components/chat/plot-in-chat.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { ImagePlot } from "@/components/plots/image";
+import { Histogram } from "@/components/plots/histogram";
+import { Linechart } from "@/components/plots/linechart";
+import { useGetPresignedUrl } from "@/hooks/get-presigned";
+import { useGetObjectFromStorage } from "@/hooks/get-storage-object";
+import { memo } from "react";
+
+interface PlotDisplayProps {
+  storageIds: string[];
+}
```
WonderPG, line 15: I think we're only using `types` in the whole code

frontend/src/hooks/get-presigned.ts
```diff
@@ -0,0 +1,18 @@
...
+      method: "GET",
+      path: "/storage/{storageId}/presigned-url",
+      pathParams: { storageId },
+    });
+    console.log(response);
+    return response as string;
+  };
+
+  return useQuery({ queryKey: [storageId], queryFn: fetchPresignedUrl });
+}
```
WonderPG, line 18: So guys this is purely for my own knowledge which I could definitely google and read on my side so feel free to ignore and resolve. I'm still pretty unfamiliar with the concepts of "hooks" so I have a couple of questions:
- What makes a hook a hook ? Is it because you used only hooks in your hooks (useXXX) ? If you just imported the fetcher and used it would it not be a hook anymore ?
- Why are hooks required here ? What is the difference between writing this as a hook or simply importing the fetcher in a utils function ? I understand that you don't want to directly fetch in the client to avoid duplication of fetch calls, hence the purpose of `react-query`, but can't you use `react-query` outside of a hook ?.

Also there's a `console.log` left here, not sure if on purpose.

frontend/src/lib/utils.ts
```diff
@@ -49,3 +55,53 @@ export function getAssociatedTools(messages: MessageStrict[]) {
...
+            viewableTools.includes(invocation.toolName) &&
+            invocation.state === "result" &&
+            invocation.result
+          ) {
+            try {
+              // Parse the result, which might be a string or an object.
+              const parsedResult = JSON.parse(
+                typeof invocation.result === "string"
+                  ? invocation.result
+                  : JSON.stringify(invocation.result),
```
WonderPG, line 91: Isn't it easier to avoid the back and forth stringify + parse if the result is an object ? i.e.
```javascript
const parsedResult = typeof invocation.result === "string"
                  ?  JSON.parse(invocation.result)
                  : invocation.result,

**Review by WonderPG — COMMENTED** [review]
Looks great thank you ! One thing I noticed though is that when you change tab the plots re-render, sometimes changing color for the pie chart

https://github.com/user-attachments/assets/76724f82-725a-49ee-937e-b551ff186f7b

Not sure if this is something we can influence. This uses the `plot generator`, not the random one so it should be deterministic.

File: frontend/src/hooks/get-presigned.ts (hunk 3 above)
Comment (BoBer78) [reply frontend/src/hooks/get-presigned.ts:18]: - A hook is function that has one `hook` inside it. If I just imported the fetcher it would have been a normal funciton yes.
- Since the react-queries are dynamic (not sure if it is the right term) --> they can be invalidated from anywhere in the app, they have to be defined in a hook.

Not 100% sure of the answer, and I will remove the `console.log()`

File: frontend/src/lib/utils.ts (hunk 4 above)
Comment (BoBer78) [reply frontend/src/lib/utils.ts:91]: This has been copied from the tools messages. It does indeed make more sense to do it that way.

**Comment by BoBer78** [comment]
> Looks great thank you ! One thing I noticed though is that when you change tab the plots re-render, sometimes changing color for the pie chart
> Screen.Recording.2025-03-21.at.11.02.02.AM.mov
> 
> Not sure if this is something we can influence. This uses the `plot generator`, not the random one so it should be deterministic.

I will set the seed to always have the same colors

**Comment by BoBer78** [comment]
Should I change how the loaders are ? for now I don't use any fancy skeletons, but I could.

**Comment by WonderPG** [comment]
If I'm not mistaken, we can remove every mention of `frontend_url` in the backend. So in config, `.env.example` and `docker-compose`

File: backend/src/neuroagent/app/dependencies.py (hunk 1 above)
Comment (BoBer78) [reply backend/src/neuroagent/app/dependencies.py]: Done in [cc82c64](https://github.com/openbraininstitute/neuroagent/pull/217/commits/cc82c64e2643a698748c85509f5c93c0afce28e6).

File: frontend/src/hooks/get-presigned.ts [hunk 5]
```diff
@@ -0,0 +1,18 @@
...
+export function useGetPresignedUrl(storageId: string) {
+  const fetcher = useFetcher();
+
+  const fetchPresignedUrl = async () => {
+    const response = await fetcher({
+      method: "GET",
+      path: "/storage/{storageId}/presigned-url",
+      pathParams: { storageId },
+    });
+    console.log(response);
```
jankrepl, line 13: I would remove

jankrepl (approved):

PR #218, opened 2025-03-18
Tool retrieval POC

### LINKED ISSUE #212: Semantic search on tools
One possible way to reduce the number of tools given to the LLM is to "preselect" them using semantic search.

The idea is to have a POC

Note that since the description of the tools can change dynamically I would not mind if we have in memory vector db that we recompute everytime the backend is started (embeddings are cheap) - but of course it would be nice to only reembed if things changed so some kind of persistance would make sense with unique hashes per content.

Inspiration

https://langchain-ai.github.io/langgraph/how-tos/many-tools/#define-the-tools

WonderPG:
Closes #212 

This PR introduces a couple of env vars, for instance:
```bash
NEUROAGENT_TOOLS__MAX_TOOLS=
NEUROAGENT_TOOLS__DEFAULT_TOOLS=
NEUROAGENT_MISC__IS_DEV=
```
will let you set the max number of tools you want (total), set the tools you always want to have (so they will be ignored by the retriever) and will assign default embeddings to everything to avoid sending requests to openai. You might need to set those when testing.


backend/src/neuroagent/app/main.py
```diff
@@ -87,6 +91,20 @@ async def lifespan(fastapi_app: FastAPI) -> AsyncContextManager[None]:  # type:
     logging.getLogger("neuroagent").setLevel(app_settings.logging.level.upper())
     logging.getLogger("bluepyefe").setLevel("CRITICAL")
 
+    # Compute tool embeddings
+    openai_client = AsyncOpenAI(api_key=app_settings.openai.token.get_secret_value())  # type: ignore
+    tool_list = get_tool_list()
+    tool_embeddings = await get_embeddings(
+        openai_client=openai_client,
+        to_embed=[tool.description for tool in tool_list],
```
jankrepl, line 99: So we only do description for now? Fine for me I guess but if this approach does not work well then this might be the reason

WonderPG: I was considering adding the title. But in theory the title should be semantically close to the description. Do you have something else in mind ?

jankrepl: For the POC it is fine. Maybe listing the parameters and their description (not recursively in case they are nested though) might be helpful? Maybe not.

PR #226, opened 2025-03-20
Rate limiting

### LINKED ISSUE #224: Implement rate limitter

jankrepl:
Closes #224 

# How to test
1. Make sure you spin up redis (you can just do `docker compose up minio redis postgres`). 
2. Use the frontend. There are basically 4 scenarios
  a) chat without vlab/project (rate limiting)
  b) chat with vlab/project (no rate limiting)
  c) suggestions without vlab/project (rate limiting)
  d) suggestions with vlab/project (no rate limiting)

3.`docker exec -it neuroagent-redis-1 redis-cli` and inspect the keys (`KEYS *`)  when then for a specific inspect the value (`GET key`) and the time to live (`TTL key`)

# FUTURE PRs
* So the toaster is only shown if we get rate limitted in the chat. We should do the same thing for the suggestions. It is just the suggestions use the fetcher under the hood and right now it does not include the request body of the error (that will contain how many seconds left to refresh) did not want to go down that rabbit hole
* Frontend error handling should rely on status codes - not random contents of response bodies

# TODO
- [x] Optional query parameters to suggestions for vlab/project + check inside they are valid for the user
- [x] Fix frontend to try to send vlab/project ID to suggestion (if present in cache)
- [x] Adjust tests for the question suggestion logic
- [x] Type annotate redis client
- [x] basic logic
- [x] Make sure toaster does not overlap with chat input
- [x] Add some minimal unit tests
- [x] Add toaster to the frontend if we are rate limted
- [x] Remove last human message if rate limited immediately
- [x] Update readme (backend and root)


docker-compose.yml
```diff
@@ -70,9 +73,19 @@ services:
...
     command: server --console-address ":9001" /data
 
+  redis:
+    image: redis:7.4.2
+    ports:
+      - "6379:6379"
+    networks:
+      - app-network
+    volumes:
+      - redis_data:/data
```
WonderPG, line 83: Is it needed to persist the data ?

frontend/src/components/chat/chat-page.tsx
```diff
@@ -153,9 +154,38 @@ export function ChatPage({
...
+
+    let errorDetail;
+    try {
+      // Try to parse error message as JSON
+      errorDetail = JSON.parse(error.message);
+    } catch {
+      errorDetail = { message: error.message };
+    }
+
+    if (errorDetail?.detail?.error === "Rate limit exceeded") {
```
WonderPG, line 174: `useChat` doesn't give you the status code of the error ?

WonderPG (commented):
Regarding the suggestions, I suggest we default to the hardcoded cards if there is an error in get suggestions. Right now everything disappears.

Otherwise LGTM thank you !
Would be great if you could just update the `.env.example`

frontend/src/components/chat/chat-page.tsx
jankrepl: This condition is indeed a bit shaky since it relies on the contents of the response body rather than the status code. I did not find a simple way how to get the the status code:(. However, I can definitely follow this PR up with an issue mentioning it

frontend/src/components/ui/toaster.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Toaster as Sonner } from "sonner";
+
+type ToasterProps = React.ComponentProps<typeof Sonner>;
+
+const Toaster = ({ ...props }: ToasterProps) => {
+  const { theme = "system" } = useTheme();
+
+  return (
+    <Sonner
+      theme={theme as ToasterProps["theme"]}
```
BoBer78, line 13: Do you need the `useTheme` ? you can maybe use the tailwind `dark:...` trick here, not sure at all if it can work

frontend/src/components/chat/chat-page.tsx
jankrepl: I actually tried that one and unfortunately the problem with it is that it is called BEFORE the `messages` are updated. So one cannot remove the last human message once you get rate limitted.

jankrepl: I mean it is for sure possible to make this nicer and again happy to create a ticket about it:)

frontend/src/components/ui/toaster.tsx
jankrepl: Great point. Honestly, I just copy pasted what the official docs suggested and was afraid to deviate

https://ui.shadcn.com/docs/components/sonner

docker-compose.yml
WonderPG: Fair, closing

frontend/src/components/chat/chat-page.tsx
WonderPG: It's a detail, if it works and there are no straightforward solution I am fine with this.

WonderPG (approved):

PR #242, opened 2025-03-27
Limit tool calls

### LINKED ISSUE #204: Set reasonable limits on number of parallel/sequential calls
Just assuming somebody is malicious and wants to explicitly do something extreme.

WonderPG:
Closes #204 

The logic goes as follow:
- You can set there two env vars
```bash
NEUROAGENT_AGENT__MAX_TURNS=
NEUROAGENT_AGENT__MAX_PARALLEL_TOOL_CALLS=
```
to control paralle/sequential tool calls
- Sequential tool calls are limited through `max_turns`. I redefined one turn as being one OpenAI call (before it was one message, i.e. 3 parallel tool calls = 4 turns). As soon as you hit `max_turns`, tool calls are deactivated, system prompt is swapped out to let the LLM know we hit rate limiting and a final call to OpenAI is made to let the user know about the limitation and ask for another message
- When getting a lot of parallel tool calls, only `max_parallel_tool_calls` are executed and the remaining ones get an answer stating that they could not be executed. The LLM is supposed to pick that up and execute them in the next turn. So we convert too many parallel tool calls into multiple turns (openAI calls) such that the `max_turns` would eventually be triggered to put a stop to this nonsense 🦫 

Of course the nice user experience relies on the LLM outputting the right messages, but the limits are hard filters and cannot be bypassed.


jankrepl (approved):
Did not try it but looks reasonable. Thank you @WonderPG

BoBer78 (dismissed):
LGTM !

jankrepl (approved):
LGTM:) Tested it locally and it works:)

PR #249, opened 2025-03-31
Extend openapi

### LINKED ISSUE #223: Remove semantic scholar tool

### LINKED ISSUE #245: Include tool outputs in openapi.json
Even if they are not part of the API (requests and responses). AFAIK there is a way to manually add arbitrary schemas to the openapi.json

This will allow frontend developers to just autogen typescript types (or even zod schemas) from the openapi specs and they won't have to guess it from observing how the JSON looks like when actual tool calling happens.

Note that not all tools spit out a JSON as an output. We also allow for `str`.

WonderPG:
Closes #245 closes #223


backend/src/neuroagent/tools/web_search.py
```diff
@@ -41,6 +40,24 @@ class WebSearchInput(BaseModel):
...
+    score: float
+    raw_content: str | None
+
+
+class WebSearchToolOutput(BaseModel):
+    """Output of the Web Search tool."""
+
+    query: str
+    results: list[SingleSearchOutput]
+    response_time: float
```
WonderPG, line 58: I couldn't find the openapi json for this API. It might exist somewhere.

jankrepl:
I tried to do `OpenAPI -> typescript` via 

```
 npx openapi-typescript http://localhost:8000/openapi.json -o generated_types.ts
```

https://github.com/openapi-ts/openapi-typescript
https://openapi-ts.dev/introduction

It works perfectly. I mean I don't want to push this solution too much since this would change the worfklow but IMO if really adopted 100% it can save a lot of time to the frontend engineers (including our frontend). For example in our `frontend/src/types.ts` a lot of the types can be simply be taken from the autogen already. I did not do this at first but I mean if you are interested feel free to POC it and add it to this PR. Again, not a must.

@WonderPG

PR #252, opened 2025-04-01
Put time in system prompt, update system prompt and make suggestions …

### LINKED ISSUE #234: Make suggestion endpoint only output 1 question
Why? The frontend only seems to be using 1 suggestion.

We should probably keep the same schema so that the frontend does not have to change anything.

### LINKED ISSUE #246: Consider adding the current time and date to system prompt
Currently we have a `Now` tool that tells us the time and date. We rely on the LLM to call it in order to fulfill requests based on relative dates.
While this is surely a valid approach, it is prone to the LLM not calling it. The model we use (`gpt-4o-mini`) thinks that we are in october 2023 so one way to be sure that it is aware of the current date could be to add it in the system prompt (compute it dynamically for each request).

Another approach could be to force the call to the `Now` tool as the first action of the LLM, OpenAI API supports that. But maybe it is an overkill

Maybe it is better to let the LLM call the `Now` tool.

Please evaluate all of the options and come up with the most efficient and simplest one.

### LINKED ISSUE #250: Update system prompt (platform description)
Based on what was given to us privately on teams.

Also, we might consider getting this "platform description" dynamically from somewhere at the backend startup (I am not sure if turning this into a tool would guarantee the LLM always uses it - similar issue like with the now tool)

WonderPG:
…return only one suggestion

Closes #234 closes #250 closes #246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Return only one suggestion.
+- Update system prompt.
+- Add current time and date in system prompt.
+
 ## [v0.4.5] - 28.03.2025
 
 ### Added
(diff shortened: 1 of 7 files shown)


jankrepl (approved):
LGTM:) Thank you :)

jankrepl (approved):

jankrepl (approved):

PR #253, opened 2025-04-01
Add rate limit headers

### LINKED ISSUE #248: Include rate limiting headers
Currently, we only have responses with status code 429 once the user gets rate limited. However, to be able to build better user experience it would be nice to return the following headers in every single rate limited endpoint we have.

* x-ratelimit-limit: Displays the applicable rate limits for the current request.
* x-ratelimit-remaining: Indicates how many requests a user can still make within the current time window.
* x-ratelimit-reset: Shows the time in seconds until the rate limit resets.

Reference: https://telnyx.com/resources/rate-limit-headers and  https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html

! Note that the idea is not only to send them when the rate limit is hit but also when it we are below the limit and return normal response with status code 200

WonderPG:
Closes #248 

I didn't feel like creating some minimal UI sorry :)

### DIFF
--- a/backend/src/neuroagent/app/app_utils.py
+++ b/backend/src/neuroagent/app/app_utils.py
@@ -4,6 +4,7 @@ import logging
 from typing import Any
 
 from fastapi import HTTPException
+from pydantic import BaseModel, ConfigDict, Field
 from redis import asyncio as aioredis
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
@@ -12,6 +13,16 @@ from starlette.status import HTTP_401_UNAUTHORIZED
 from neuroagent.app.config import Settings
 
 logger = logging.getLogger(__name__)
+
+
+class RateLimitHeaders(BaseModel):
+    """Headers for the rate limits."""
+
+    x_ratelimit_limit: str = Field(alias="x-ratelimit-limit")
+    x_ratelimit_remaining: str = Field(alias="x-ratelimit-remaining")
+    x_ratelimit_reset: str = Field(alias="x-ratelimit-reset")
+
+    model_config = ConfigDict(populate_by_name=True)
 
 
 def setup_engine(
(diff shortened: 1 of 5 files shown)


backend/src/neuroagent/app/app_utils.py
```diff
@@ -14,6 +15,16 @@
...
 
 
+class RateLimitHeaders(BaseModel):
+    """Headers for the rate limits."""
+
+    x_ratelimit_limit: str = Field(alias="x-ratelimit-limit")
+    x_ratelimit_remaining: str = Field(alias="x-ratelimit-remaining")
+    x_ratelimit_reset: str = Field(alias="x-ratelimit-reset")
+
+    model_config = ConfigDict(populate_by_name=True)
```
jankrepl, line 25: Seems like this one will be deprecated soon https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.populate_by_name

`validate_by_name` should behave the same

backend/src/neuroagent/app/routers/threads.py
```diff
@@ -77,17 +77,24 @@ async def generate_title(
     settings: Annotated[Settings, Depends(get_settings)],
     thread: Annotated[Threads, Depends(get_thread)],
     redis_client: Annotated[aioredis.Redis | None, Depends(get_redis_client)],
+    response: Response,
```
jankrepl, line 80: quite confusing because we use the `response` as the output for openai...I am surprised mypy does not yell

jankrepl, line 80: quite confusing because we use the `response` as the output for openai...I am surprised mypy does not yell

backend/src/neuroagent/app/app_utils.py
WonderPG: Good catch. Funny they made a mistake in the doc
image

Anyway fixed in 581ae3a

backend/src/neuroagent/app/routers/threads.py
WonderPG: Fixed in 581ae3a

jankrepl (approved):
LGTM:)) Thanks

PR #261, opened 2025-04-04
Filter and sort

### LINKED ISSUE #10: Get threads / messages : allow filtering and sorting
get threads: sort query parameter (allow to sort based on any column, &sort=creation_date , &sort=-creation_date, ...

WonderPG:
Closes #10

### DIFF
--- a/backend/src/neuroagent/app/routers/threads.py
+++ b/backend/src/neuroagent/app/routers/threads.py
@@ -7,7 +7,7 @@ from typing import Annotated, Any, Literal
 from fastapi import APIRouter, Depends, HTTPException, Query, Response
 from openai import AsyncOpenAI
 from redis import asyncio as aioredis
-from sqlalchemy import desc, or_, select
+from sqlalchemy import desc, or_, select, true
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
 
@@ -126,7 +126,7 @@ async def get_threads(
     project_id: str | None = None,
     sort: Literal[
         "update_date", "creation_date", "-update_date", "-creation_date"
-    ] = "update_date",
+    ] = "-update_date",
 ) -> list[ThreadsRead]:
     """Get threads for a user."""
     validate_project(
@@ -143,9 +143,9 @@ async def get_threads(
             Threads.project_id == project_id,
         )
         .order_by(
-            getattr(Threads, sort_column)
+            desc(getattr(Threads, sort_column))
             if sort.startswith("-")
-            else desc(getattr(Threads, sort_column))
+            else getattr(Threads, sort_column)
         )
     )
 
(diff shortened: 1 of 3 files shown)


backend/src/neuroagent/app/routers/threads.py
```diff
@@ -124,17 +124,29 @@ async def get_threads(
     user_info: Annotated[UserInfo, Depends(get_user_info)],
     virtual_lab_id: str | None = None,
     project_id: str | None = None,
+    sort: Literal[
+        "update_date", "creation_date", "-update_date", "-creation_date"
```
BoBer78, line 128: I am not sure I like the convention of having a minus to indicate a descending order. Definitely a nitpick.

backend/src/neuroagent/app/routers/threads.py
```diff
@@ -124,17 +124,29 @@ async def get_threads(
...
+    ] = "-update_date",
```
jankrepl, line 129: This is fine but just note one way to generalize this is to allow for a list of these. That way you can sort on multiple columns at the same time (would matter in cases where the first sorting column has multiple identical values)

Anyway, probably an overkill for now. Maybe something for the future

backend/src/neuroagent/app/routers/threads.py
jankrepl: Fair. AFAIK as with many other REST related concepts there is no fixed convention how to do this.

What I have seen
1. `sort=field` and `sort=-field`
2. `sort=field desc` and `sort=field asc` (the space would have to be url encoded)
3. `direction=desc&sort=field` and  `direction=asc&sort=field`. The problem with this one you can only sort by 1 field at a time. Anyway, the PR does not support multiple sort fields anyway.


https://stackoverflow.com/questions/578380/url-query-string-convention-for-multiple-sort

backend/src/neuroagent/app/routers/threads.py
```diff
@@ -200,18 +212,24 @@ async def get_thread_messages(
...
+        entity_where = true()
     messages_result = await session.execute(
         select(Messages)
-        .where(
-            Messages.thread_id == thread_id,
-        )
+        .where(Messages.thread_id == thread_id, entity_where)
         .options(joinedload(Messages.tool_calls))  # Eager load tool_calls
-        .order_by(Messages.order)
+        .order_by(desc(Messages.order) if sort.startswith("-") else Messages.order)
```
jankrepl, line 232: I mean we might as well directly sort on the creation date no? Given the fact we will get rid of the `order` attribute.
As discussed, (unless we missed something) the two approaches should lead to identical sorting.

Or am I missing something. Anyway, we might also do it in a different PR. It is just a bit misleading to pass a query parameter `sort=creation_date` and then sort on a different column

backend/src/neuroagent/app/routers/threads.py
WonderPG: Fair, in theory you will almost never have two threads with the same creation date but sure to keep in mind

jankrepl (approved):

PR #267, opened 2025-04-11
Plot traces

### LINKED ISSUE #169: Plot traces tool
**Requirements**
* Should work for both in vitro and in silico traces
* It should probably be multiple line charts overlayed (corresponding to different input stimuli)
  * You decide whether the plotting takes place on the server (matplotlib) or client (chart.js) - but if you go for client maybe this will require a new plot time because the current plot types don't support multiple lines
https://github.com/openbraininstitute/neuroagent/blob/52b12a87589a6b9561e56041bcbfd9db3f74516e/backend/src/neuroagent/schemas.py#L16-L23
* Probably also get inspiration from how the core web app plots things

**Extra**
Maybe we should make it possible to have it "streamed" - like when the simulation is running 

Feel free to ping @jankrepl when it comes to the storage/plotting logic

BoBer78:
closes #169


backend/src/neuroagent/app/dependencies.py
```diff
@@ -167,7 +167,8 @@ def get_tool_list() -> list[type[BaseTool]]:
     """Return a raw list of all of the available tools."""
     return [
         SCSGetAllTool,
-        SCSGetOneTool,
+        # SCSGetOneTool,
```
WonderPG, line 170: So the `GetOneTool` is not used anymore ? What if we want to get a simulation without necessarily plotting it ?

backend/src/neuroagent/tools/bluenaas_scs_plot.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class SCSPlotTool(BaseTool):
+    """Class defining the SCSGetOne tool."""
+
+    name: ClassVar[str] = "scsplot-tool"
+    name_frontend: ClassVar[str] = "Plot Single-Neuron Simulation"
+    description: ClassVar[str] = """Make a plot of the simulation, based on its id.
+    The id can be retrieved using the 'scs-getall-tool', from the simulation report of `scs-post-tool` or directly specified by the user.
+    The plots will be shown after / below your message automatically.  Use the simulaton label to refer to the the plots. DO NOT MENTION the storage id nor the simulation ID. DO NOT EMBED the plots in the text."""
```
WonderPG, line 51: Double `the`

backend/src/neuroagent/tools/bluenaas_scs_plot.py
```diff
@@ -0,0 +1,123 @@
...
+    description_frontend: ClassVar[
+        str
+    ] = """Make a plot from the result of your simulation.
```
WonderPG, line 54: I would specify "single cell simulation"

backend/src/neuroagent/tools/bluenaas_scs_plot.py
```diff
@@ -0,0 +1,123 @@
...
+        identifiers = [
+            save_to_storage(
+                s3_client=self.metadata.s3_client,
+                bucket_name=self.metadata.bucket_name,
+                user_id=self.metadata.user_id,
+                content_type="application/json",
+                category=plot.category,
+                body=plot.model_dump_json(),
+                thread_id=self.metadata.thread_id,
+            )
```
WonderPG, line 111: Shoudn't this function be async ? I guess we have a bottleneck here. Not really related to this PR though.

WonderPG (commented):
I could not really test it because I kept getting Nexus errors, but lgtm beside the small comments

backend/src/neuroagent/app/dependencies.py
BoBer78: We can re-activate this tool, from what remember, appart from the x,y coordinates this tool did not give much more information than the `get-all` tool.

backend/src/neuroagent/tools/bluenaas_scs_plot.py
```diff
@@ -0,0 +1,123 @@
...
+    bucket_name: str
+    thread_id: str
+
+
+class InputSCSPlotTool(BaseModel):
+    """Inputs for the Bluenaas plot tool."""
+
+    simulation_id: str = Field(
+        description="ID of the simulation to retrieve. Should be an https link."
+    )
```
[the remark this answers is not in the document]
BoBer78: We could let the LLM choose some plot parameters, I wanted to keep it simple. We could for example let it choose the tile and the color of the lines like the other plots we have.

WonderPG (approved):
🦫

PR #270, opened 2025-04-14
Remove sqlite support

### LINKED ISSUE #255: Don't support sqlite
Be careful about the "unit"/integration tests that use the DB. Maybe create a database inside the postgres where the tests will be pointing at.

We are hoping that this will make our life easier and the migration scripts will have to work only for postgres.

Where to change?
* Pytest mark parametrize
* simplify existing migration scripts
* simplify alembic settings
* github action

WonderPG:
Closes #255


jankrepl (approved):
Great!! Thank you

PR #271, opened 2025-04-14
Remove order column from the messages table

### LINKED ISSUE #78: Remove `order`  column in messages and order by creation date
Ordering by creation date is virtually as powerful as ordering by `order`.  Another benefit is that when you need to insert a new message in the DB without having the previous list of message, you don't need to do
```python
 latest_message = await session.execute(
        select(Messages)
        .where(Messages.thread_id == thread_id)
        .order_by(desc(Messages.order))
        .limit(1)
    )
    latest = latest_message.scalar_one_or_none()
    new_message.order = (latest.order + 1) if latest else 0
```
since the creation date is automatically "the latest". Also it creates itself.

WonderPG:
Closes #78

### DIFF
--- a/backend/src/neuroagent/app/routers/qa.py
+++ b/backend/src/neuroagent/app/routers/qa.py
@@ -228,7 +228,6 @@ async def stream_chat_agent(
                 )
                 messages.append(
                     Messages(
-                        order=len(messages),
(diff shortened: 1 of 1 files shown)


jankrepl (approved):
~Did not find any issues:) Thank you!~

PR #277, opened 2025-04-15
fix LS litterature + small bug in guardrail

### LINKED ISSUE #113: Literature search URL env var should point at the root (not `retrieval/` endpoint)
In general, when we provide URLs as env vars...let's make sure they point at the base of the API .

Why?
Well, currently we want to not only ping a specific endpoint but also ping a healthcheck endpoint. Writing these random suffix striping if statements is very annoying

`BASE_URL/retrieval` vs `BASE_URL`.

I would follow this convention with ALL external APIs we have.

BoBer78:
closes #113


WonderPG (approved):
LGTM thank you !

Issue #278, opened 2025-04-15
Semantic router: False positives


jankrepl:
For now this ticket serves as a place where we gather false positives we encountered. We can act on them later by

1. Increase the threshold - fewer false positives
2. Create a new semantic route that corresponds to past false positives and just keep on appending utterances to it ..and if we hit it we just trigger the standard process

# False positives from the past (feel free to append
* Can you please find a me model and run a simulation on it?

WonderPG:
Dropping some stats here:
```
* Can you please find a me model and run a simulation on it?
[('Human data',
  0.3792532945309211,
  [0.3068230666366712,
   0.33058058010632596,
   0.4232388556637351,
   0.45637067571695217]),
 ('Agent capabilities', 0.30086823064239776, [0.30086823064239776])]

* List me some me models
[('Human data', 0.34957454785628544, [0.3109130580181943, 0.3882360376943766]),
 ('Agent capabilities',
  0.2568437899101416,
  [0.24141018469252448, 0.28485000443964925, 0.24427118059825098])]

* can you please make a plot of my simulation ?
[('Human data', 0.2866082352085344, [0.2627915507045384, 0.31042491971253044]),
 ('Agent capabilities',
  0.2682845662772633,
  [0.26949068361017925, 0.2657730699108199, 0.2695899453107907])]

* How many in total are available ?
[('Agent capabilities',
  0.3024888173279417,
  [0.26589429511750184, 0.2788885797682874, 0.36268357709803584]),
 ('Human data', 0.2905660380335009, [0.2798423154898038, 0.30128976057719803])]

* Show me papers about intracellular patch clamp data for the thalamus of rodents
[('Human data',
  0.3034044720118845,
  [0.19851474215122084, 0.5169340744854313, 0.1947645993990012]),
 ('Agent capabilities',
  0.1930241933288508,
  [0.16943749058748742, 0.2166108960702142])]

* Get me a list of ME models

[('Human data', 0.33315313821669096, [0.3084252375396819, 0.35788103889370004]),
 ('Agent capabilities',
  0.23929386461818478,
  [0.21313720579860718, 0.24157718517838248, 0.2631672028775647])]


GOOD TRIGGERS
 What kinds of things can you help me with?
[('Agent capabilities',
  0.4699211014220168,
  [0.3010576640023665,
   0.5772200096797835,
   0.47982177501410134,
   0.568149966443398,
   0.42335609197043456])]

Tell me about your functions and abilities.
[...]

PR #279, opened 2025-04-15
Update fixed_llm_responses.yml

danielaegassan:
I'm not totally sure I have the right format for this suggestion, so let me know if I need to do any changes.  This should help us address the human data issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Basic Guardrails.
 - Possibility to interrupt streaming.
 - Filter and sort get_thread and get_thread/messages.
+- New semantic route.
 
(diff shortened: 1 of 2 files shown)


WonderPG (approved):
@danielaegassan The format is great, it works as expected 
image
Thanks for trying it and feel free to merge when you want. 

Two things to note:
- These changes won't take effect before we deploy to staging (i.e. not before next week most likely. Ping us if you would like them out before)
- False positives **WILL** happen. We will try to mitigate them but to do that we need to gather them. We opened this ticket https://github.com/openbraininstitute/neuroagent/issues/278 where you can give us the queries that triggered a false positive. Please if you encounter one, write it there.

Thanks !

jankrepl (approved):

WonderPG:
@danielaegassan exactly. The fixed response triggers when you didn't expect it to trigger.

PR #280, opened 2025-04-16
Frontend fixes

### LINKED ISSUE #134: Nonexistent tools from older conversations cause error
If you create and invoke a tool in one thread and then remove the tool from the backend and try to reopen that conversation you get an error.

It is because of this code

https://github.com/openbraininstitute/neuroagent/blob/27b9b3b498a8c792f816151dafe2f52da9e089fd/frontend/src/components/chat-message-tool.tsx#L167-L173

TLDR: the available tools are taking from the current backend source code, whereas tool name is read from the db.

I would just extend the code to default to the slug rather than the `name_frontend` in case the available tools don't contain it

I think this is the first case of a problem we will hit more often: there are 2 sources of truth
* the source code (e.g. HIL, `name_frontend`,)
* DB

Is this problem artificial? It depends. If you are switching between different branches and they don't have the same tools in the backend you hit it

### LINKED ISSUE #201: Dynamic metadata (title)
Currently, we have the same static metadata for every single page of our application.

https://github.com/openbraininstitute/neuroagent/blob/7ca8eccf84ea7cafce9512f133bcf7dcbbe8ab0e/frontend/src/app/layout.tsx#L25-L28

However, nextjs makes it quite simple to generate the metadata (including the title) in a dynamic way - https://nextjs.org/docs/app/building-your-application/optimizing/metadata#dynamic-metadata.  I suggest that we make sure the below pages have dynamic **titles** (what is shown in the browser tab) (the other pages can have the current static metadata)

* `/threads/{threadId}` - display thread title (possibly truncated)
* `/tools/{toolSlug}` - display frontend tool name (possibly truncated)

It is useful if you have multiple tabs open (among other things). Note that it should not slow down the existing logic - ideadlly the fetch calls should be cached - please double check

Note that chat gpt does it too 

https://github.com/user-attachments/assets/b1fba5ac-ab45-4bf9-b5d5-2f1787feeb6b

### LINKED ISSUE #230: Don't mix default/nondefault exports + always use absolute imports
https://github.com/openbraininstitute/neuroagent/blob/d31a9bc5be030adc1ba6675029687e80bcc448f6/frontend/src/components/chat/chat-input.tsx#L3-L15

Please research the best practices online. I created this issue because of subjective opinions and I am happy to go for the best practices. Also, if there is a eslint/prettier plugin enforcing some of this let's go for it.

* Let's not mix relative "./...." and absolute "@/.." imports
* At the start no default exports were used in components, but it seems like now sometimes we use them   (e.g. ./question-suggestion-cards")

BoBer78:
closes #134 closes #201 closes #230


frontend/src/app/viewer/[storageId]/page.tsx
```diff
@@ -6,6 +6,12 @@ import { ImagePlot } from "@/components/plots/image";
 import { Histogram } from "@/components/plots/histogram";
 import { Linechart } from "@/components/plots/linechart";
 
+export async function generateMetadata() {
+  return {
+    title: "Plot viewing",
```
Comment (WonderPG) [inline frontend/src/app/viewer/[storageId]/page.tsx:11]: Is it possible to use the plot title here ?

WonderPG (commented):
Tab title is missing in the following cases:
- The url is incorrect (error)
- Settings page

frontend/src/app/viewer/[storageId]/page.tsx
Comment (BoBer78) [reply frontend/src/app/viewer/[storageId]/page.tsx:11]: I could, but I have to wait to fetch the pre-signed URL, and this is done on the client. Not sure if there is a smart way to do it

BoBer78:
> Tab title is missing in the following cases:
> 
>     * The url is incorrect (error)
> 
>     * Settings page

This was fixed in [1f504ea](https://github.com/openbraininstitute/neuroagent/pull/280/commits/1f504ea76dedfbc560278d95c2ad8fd62141f156)

PR #285, opened 2025-04-24
Increase thresholds for semantic router

### LINKED ISSUE #282: Adjust semantic router threshold
Currently very obviously wrong queries trigger the semantic router. I suggest as a temporary fix we increase a bit the threshold to a somewhat reasonable value. We have a couple of examples in #278 that can be used to tweak it a bit. This is most likely temporary but I can already see the thousands of complaints we will have as soon as we release.

WonderPG:
Closes #282

Temporary fixes to avoid complaints


BoBer78 (dismissed):
🦫

BoBer78 (approved):
🦫 🪜 🍩

PR #288, opened 2025-04-28
Scite Ai test

### LINKED ISSUE #287: Experiment with scite AI API
We have access till the 5th of may to their service. We apparently have access to the snippets but cannot use it in a RAG application. Not too sure what this means but the idea would be to experiment with their API, find the right endpoints (most likely the search ones) and to eventually create a new tool from it. 

Once we have it I assume we can do a demo to the management but **Do not deploy this change to staging/prod**.

BoBer78:
closes #287


WonderPG:
Closing as we won't go for Scite

PR #291, opened 2025-04-30
Fix sandbox issues

### LINKED ISSUE #220: Create target group for minio and make sure client -> minio works
So with #160 we will need to expose our sandbox minio via a target group and make sure that the presigned url actually has the correct address.

IMO this is actually trickier than it might seem: https://github.com/minio/minio/issues/8007 

Also, we need to make sure the minio UI is not accessible easily (e.g. custom pwd and user)

### LINKED ISSUE #262: Agent image not buildable on arm platform
We have moved to python 3.11 for the accounting. For some reason morphio is not able to be compiled on `arm` platforms. This is an issue with the version of Cmake. I was not able to fix it friday so I open an issue to keep it in mind. 
(Everything works fine on amd64 platforms)

jankrepl:
Closes #262 and closes #220 

- [x] Make sure presigned url (client -> minio) has a different URL then fastapi backend -> minio
- [x] Build Morphio from scratch

So for the first point, I used the concept of a network `alias` that one can add in `docker-compose` and it basically corresponds to a DNS entry that the given service will use for resolving.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -2,12 +2,26 @@ FROM python:3.11
 
 ENV PYTHONUNBUFFERED=1
 
-RUN apt-get -y update
-RUN apt-get -y install curl \
-    cmake \
+RUN apt-get -y update && \
+    apt-get -y install \
+    curl \
     build-essential \
     libhdf5-dev \
-    ninja-build
+    ninja-build \
+    wget \
+    && rm -rf /var/lib/apt/lists/* \
+    && wget https://github.com/Kitware/CMake/archive/refs/tags/v4.0.1.tar.gz -O cmake.tar.gz \
+    && tar xzf cmake.tar.gz \
+    && cd CMake-4.0.1 \
+    && ./bootstrap --parallel=$(nproc) \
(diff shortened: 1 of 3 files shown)


WonderPG (approved):

backend/Dockerfile
```diff
@@ -2,12 +2,26 @@ FROM python:3.11
...
+    && cd CMake-4.0.1 \
+    && ./bootstrap --parallel=$(nproc) \
+    && make -j$(nproc) \
+    && make install \
+    && cd .. \
+    && rm -rf CMake-4.0.1 cmake.tar.gz
+
+ENV CMAKE=/usr/local/bin/cmake
+ENV CMAKE_POLICY_VERSION_MINIMUM=3.5
+RUN which cmake && cmake --version
```
BoBer78, line 24: Is there not a way to check if we are on and `arm` platform before running the extra logic ?

jankrepl: We agreed to see how fast it is in our CI (when we have a new tag) and if it is slow I will create a new ticket and fix it.

PR #292, opened 2025-04-30
Question suggestions in chat

### LINKED ISSUE #256: Suggestions within a thread
Very related to #13 .

The minimal way to implement this would be to add a new column to the `threads` table called `summary` and it would keep the running summary of the whole conversation (that means it will have to be updated whenever we evoke chat_streamed - see the sketch in the linked issues) and then we would create a new suggestion endpoint `POST /threads/{thread_id}/question_suggestions`  that would take the `summary` of the thread and send it over to an LLM (structured output) to get some suggestions

BoBer78:
closes #256


backend/src/neuroagent/app/routers/qa.py
```diff
@@ -131,6 +139,113 @@ async def question_suggestions(
...
+            virtual_lab_id=vlab_id,
+            project_id=project_id,
+        )
+        limit = settings.rate_limiter.limit_suggestions_inside
+    else:
+        limit = settings.rate_limiter.limit_suggestions_outside
+
+    limit_headers, rate_limited = await rate_limit(
+        redis_client=redis_client,
+        route_path="/qa/question_suggestions_in_chat",
```
WonderPG, line 168: By having a `route_path` different than the other suggestion one, we separate the two counts of suggestions, i.e. the suggestion for the first message doesn't count toward the suggestions after the first message and vice versa. It is a design choice which I'm not necessarily against, but just pointing it out.

WonderPG: Also in theory (I insist on the theory) a person sends more messages than he/she starts threads. It would in this case make sense to have different rate limits for `suggestion_first_message` and `suggestion_in_thread`. But I don't like adding more env vars here. I wonder if having one global `suggestion` count/limit is not better.

backend/src/neuroagent/app/routers/qa.py
```diff
@@ -131,6 +139,113 @@ async def question_suggestions(
...
+            "Users can also customize the models or create their own ones and change the cellular composition, and then run simulation experiments and perform analysis. "
+            "The user is navigating on the website, and we record the last elements he accessed on the website. Here is what the user's history will look like :"
+            "user_history = [[['brain_region', 'example'], ['artifact', 'example'], ['artifact', 'example'], ['artifact', 'example']], [['brain_region', 'example'], ['artifact', 'example']]]"
+            "'brain_region' can be any region of the mouse brain."
+            "'artifact' can be :  'Morphology','Electrophysiology','Neuron density','Bouton density','Synapse per connection','E-model','ME-model','Synaptome' "
+            "and 'data_type' can be 'Experimental data' or 'Model Data'"
+            "The last element of the list represents the last click of the user, so it should naturally be more relevant."
+            "From the user history and previous messages, try to infer the user's intent on the platform. The messages have a lot more importance than the user journey."
```
WonderPG, line 217: I don't think we are trying to infer the user's intent on the platform here, rather the next question he might want to send to the chat.

backend/src/neuroagent/app/routers/qa.py
```diff
@@ -131,6 +139,113 @@ async def question_suggestions(
...
+            "From it generate some questions the user might want to ask to a chatbot that is able to search for papers in the literature."
+            "The questions should only be about the literature. Each question should be short and concise. In total there should three questions.",
+        },
+        {
+            "role": "user",
+            "content": "Here is the users journey : \n"
+            + json.dumps(body.click_history)
```
WonderPG, line 224: Isn't `body.model_dump_json()` enough ?

backend/src/neuroagent/app/routers/qa.py
```diff
@@ -131,6 +139,113 @@ async def question_suggestions(
...
+                        "role": "user"
+                        if msg.entity == Entity.USER
+                        else "assistant",  # since we filtered out the rest
+                        "content": json.loads(msg.content)["content"],
+                    }
+                    for msg in db_messages
+                ]
+            ),
+        },
+    ]
```
WonderPG, line 238: The `content` field in the DB already contains the `role` field, here is a sample `content` from the DB:
```
{
  "role": "user",
  "content": "Find the id of the thalamus"
}
```
So a simple `json.dumps(msg)` should do the trick.

WonderPG (commented):
Thank you 🦫

backend/src/neuroagent/app/routers/qa.py
BoBer78: Yeah, since the endpoint will be barely used, no need to complicate things. Besides, it will be strange to have suggestions in chat active and not the one on startup (there will still be the ones from Fabien)

backend/src/neuroagent/app/routers/qa.py
BoBer78: Well it is the difference between : 
```
`[[["Auditory cortex"], ["Douroum brain"]]]`
```
or 
```
'{"click_history":[[["Auditory cortex"],["Douroum brain"]]]}'
```

PR #298, opened 2025-05-09
merge two suggestion endpoints

BoBer78:


WonderPG (commented):
Instead of having a bunch of conditional statements, I recommend we always use one pydantic class that always outputs 3 suggestions. In the prompt we state that we "pass the click history and the conversation history which might be empty." Then we dump both, with the history potentially being an empty list. Then the frontend can chose 1 or 3 outputs as they need.

jankrepl (dismissed):
LGTM:)

jankrepl (approved):

PR #299, opened 2025-05-09
Integrate morpho metrics with obi one

### LINKED ISSUE #284: Integrate `morpho_feature` tool with obi-one
We have to wait for the green light from James but we have been tasked to call the endpoint he is integrating into obi-one to compute morphology features. He still has a couple of things to change though.

The plan is to help him make a very clean first endpoint so that scientists can use it as a template. The input parameters should be very well documented on his side so that we can use an openapi.json to pydantic tool to immediately get the input/output classes. This ticket should be very minimal work, simply change our tool to making an HTTP request to his deployed endpoint. Every downstream issue should be reported to him.

WonderPG:
Closes #284


james-isbister (commented):
Looks good!

jankrepl (approved):
Had a look at the code and looks great to me!!

jankrepl (approved):

PR #312, opened 2025-05-20
Adapt autogenerated types

### LINKED ISSUE #309: Migrate existing autogenerated types to the new tool
The goal here is to generate `models.py` + `schema.py` with the new auto generation tool, and use those new types in existing tools. It should be done for:
- obi-one
- literature search
- bluenaas
and underlying tools should be adapted.

### LINKED ISSUE #310: Pydantic <-> mypy does not work
Obvious typing mismatches in Pydantic BaseModel are not picked up by when inside `src` and outside `src` - might be related to the plugin.

It has to do with the pydantic-mypy plugin

```
[tool.pydantic-mypy]
init_forbid_extra = true
init_typed = true # <------- THIS IS THE ISSUE
warn_required_dynamic_aliases = true
```

```
from pydantic import Field, BaseModel, field_validator

class Parent(BaseModel):
    a: int = 1


# @field_validator("a")
# def check_a(cls: 'Parent', v: str | int) -> str | int:
#     if isinstance(v, str):
#         return int(v)
#     return v

parent = Parent(a="5325235235")


```

WonderPG:
Closes #309 closes #310


backend/src/neuroagent/tools/bluenaas_memodel_getall.py
```diff
@@ -1,19 +1,31 @@
...
 logger = logging.getLogger(__name__)
 
 
+class InputMEModelGetAll(
+    RetrieveNeuronModelsApiBluenaasNeuronModelVirtualLabIdProjectIdMeModelsGetParams
+):
+    """Inputs for the BlueNaaS single-neuron simulation."""
+
+    virtual_lab_id: SkipJsonSchema[str] = "placeholder"
+    project_id: SkipJsonSchema[str] = "placeholder"
```
WonderPG, line 26: This pattern (i.e. annotation with `SkipJsonSchema`) means "take all from the parent class but do not expose those to the LLM"

backend/src/neuroagent/tools/morpho_metrics_tool.py
```diff
@@ -26,8 +34,8 @@ class MorphoMetricsMetadata(BaseMetadata):
     token: str
     httpx_client: AsyncClient
     obi_one_url: str
-    vlab_id: str | None
-    project_id: str | None
+    vlab_id: UUID | None
+    project_id: UUID | None
```
WonderPG, line 38: API requirements from obi-one request a UUID, while in other APIs it is a string. Maybe something that we should tell James.

jankrepl (approved):
LGTM:)))

Thanks @WonderPG

PR #315, opened 2025-05-22
Entitycore resolving

### LINKED ISSUE #306: Implement custom brain region resolving
We got the blessing to do this ourselves since the `ilike` on the entitycore endpoints is not powerful enough.

Consider actually implementing this feature on the `entitycore` side.

Possible solutions:
1. Laziest: Make sure the brain region has the tsvector assigned to it (based on the name) 
 * https://github.com/openbraininstitute/entitycore/blob/8b434a79c8f30a57d4210ca93dcfcfcdf2b85abc/app/db/model.py#L101-L110 
 * https://github.com/openbraininstitute/entitycore/blob/8b434a79c8f30a57d4210ca93dcfcfcdf2b85abc/app/db/model.py#L325-L332

WonderPG:
Closes #306 

# How to test
Get your (staging) token and run 
```bash
python src/neuroagent/scripts/embed_hierarchies.py $token -e "https://staging.openbraininstitute.org/api/entitycore"
```
Not providing arguments will use the variables present in your environment. Make sure Minio is running and your environment variables are set.

The above will add a file in your minio storage which contains pre-computed embeddings for every single brain region name of the default hierarchy.

Then simply start your app and test the tool.


backend/src/neuroagent/scripts/embed_hierarchies.py
```diff
@@ -0,0 +1,187 @@
...
+    (
+        parser.add_argument(
+            "--s3-secret-key",
+            "-s",
+            type=str,
+            required=False,
+            default=None,
+            help="Secret key of the s3 bucket. Read from env if not specified.",
+        ),
+    )
```
jankrepl, line 86: Not saying it is wrong. But IMO you made this even more complicated than it needs to be.


What you did is `entitycore -> s3`. Alternatively, what we can do is `entitycore -> local file` and then let the user independently put the file in minio/s3. It is quite easy using AWS CLI or Minio CLI.


See our current readme that shows how to do `local file -> minio` 

https://github.com/openbraininstitute/neuroagent/blob/77df05682dd9970d621ef60780efb3de8f104232/README.md?plain=1#L26-L27




Anyway. As you wish. We can also keep the current logic.

WonderPG: The whole logic is broken if you don't upload the file to the storage. The startup also loads it from minio without giving an option for local files. So if you don't take the extra step and run the command above, the tool won't work. So either we try to make it accessible from local file in the whole code, or we enforce it to be on minio/s3 imo. What do you think ?

WonderPG: Also note that we made the decision to make minio part of the infrastructure as a required component.

backend/src/neuroagent/schemas.py
```diff
@@ -149,3 +149,19 @@ class JSONMultiLinechart(BaseObject):
...
     )
     line_color: str | None = Field(None, description="Hex color code for the line")
+
+
+class EmbeddedBrainRegion(BaseModel):
+    """Brain region schema."""
+
+    id: str
+    name: str
+    hierarchy_level: int
```
jankrepl, line 159: Why do we need this `hierarchy_level`? Feels a bit arbitrary and from what I have seen the tool itself does not use it.

WonderPG: Just a future proof thing. I'm expecting that we will eventually prompt engineer the LLM to chose BRs from a lower hierarchy level for underlying tool calls. Since the hierarchy is resolved on entity core side, having a higher up hierarchy will embed the lower ones. Maybe premature, but I expect this to come soon.

WonderPG: I've heard this idea fly by a couple of times.

jankrepl: Ok:) We will see what happens.

backend/src/neuroagent/scripts/embed_hierarchies.py
jankrepl: I see your point. Some counter arguments.


1) As discussed privately though, we might not have write access to prod/staging S3. So the script won't work anyway.
2) Using the CLI is way simpler than all the Python boto3 code you had to write. 


Anyway, I am fine with going for your solution but please write in the readme explicitly that if one starts `docker compose` from scratch and the volumes are empty you need to run the script.

WonderPG: Sure I will add it to the README. Regarding your points:
1. Sure, for prod it won't work but the effort to get the file in prod is the same regardless
2. I would argue it's the opposite: my code
```python 
s3_client = boto3.client(
        "s3",
        endpoint_url=s3_url or os.getenv("NEUROAGENT_STORAGE__ENDPOINT_URL"),
        aws_access_key_id=s3_access_key or os.getenv("NEUROAGENT_STORAGE__ACCESS_KEY"),
        aws_secret_access_key=s3_secret_key
        or os.getenv("NEUROAGENT_STORAGE__SECRET_KEY"),
        aws_session_token=None,
        config=boto3.session.Config(signature_version="s3v4"),
    )

    s3_client.put_object(
        Bucket=s3_bucket_name or os.getenv("NEUROAGENT_STORAGE__BUCKET_NAME"),
        Key=f"shared/{hierarchy_id}_hierarchy_embeddings.json",
        Body=brain_regions.model_dump_json(),
        ContentType="application/json",
    )
```
plus the CLI part, which is very basic argparse.
The command
```
docker cp brainregion_hierarchy.json neuroagent-minio-1:/tmp/brainregion_hierarchy.json && docker exec -it neuroagent-minio-1 mc alias set myminio http://minio:9000 minioadmin minioadmin && docker exec -it neuroagent-minio-1 mc cp /tmp/brainregion_hierarchy.json myminio/neuroagent/shared/brainregion_hierarchy.json 
```
I find it way more complex 😅

jankrepl: Discussed privately. Current solution is fine we just add a small note in the readme.

WonderPG: Done in 3056b6c

PR #317, opened 2025-05-23
Centralize bearer token in dependency

### LINKED ISSUE #316: Inject authentication header directly in the dependency
In our app we share the httpx client among all of the tools and more. The dependency is defined as follow:
```python
async def get_httpx_client(request: Request) -> AsyncIterator[AsyncClient]:
    """Manage the httpx client for the request."""
    client = AsyncClient(
        timeout=300.0,
        verify=False,
        headers={"x-request-id": request.headers["x-request-id"]},
    )
    try:
        yield client
    finally:
        await client.aclose()
```
We can insert directly the authorization header here by depending on the auth (i.e. add `token: Annotated[str, Depends(auth)] in the dependency's input) and propagate it this way through the entire app. We should also remove it from every single place that uses this httpx client.

### LINKED ISSUE #331: Remove `httpx_client` from `BaseMetadata`
https://github.com/openbraininstitute/neuroagent/blob/9ff2c745bf23dfdb5ac485d501c41fe4ebdf2028/backend/src/neuroagent/tools/base_tool.py#L60

Make sure to add it in the tool specific metadata subclasses (only those who actually need it)

**Why?** 
Not all tools need an HTTPX client

WonderPG:
Closes #316 closes #331


jankrepl (approved):
LGTM:)


Let's just make sure not to use that client for 3rd party API calls

PR #318, opened 2025-05-23
Update use chat

### LINKED ISSUE #186: Update to newest vercel ai sdk (useChat)
So it seems like they introduced a lot of new features (e.g. reasoning,  the `useChat`'s `messages` has a different type,  the underlying data streaming protocol changed,...)

In our lock file we have `4.0.37`, however, the latest version (and the one that the online docs refer to) is `4.1.54`. So if we believe their semantic versioning logic there should be no backward breaking changes. But let's please double check that and at the same time migrate to whatever the latest types are etc.

Unlike other tools it seems like the docs of vercel ai sdk don't give you the option to see the docs of older versions (maybe possible, I just did not find it) then feel free to just serve the docs locally (I guess it should be possible) so that it is easier to understand the changes. Or maybe check their changlog

BoBer78:
closes #186


jankrepl:
It works great. I did not find any issues during my testing. Of course this PR is huge and there are so many details to look into. It is just that I am fine with merging and then potentially fixing bugs later.

PR #322, opened 2025-05-26
Fix CI

jankrepl:
The issues comes from the new version of `semantic-router` - https://pypi.org/project/semantic-router/0.1.8/


BoBer78 (approved):
LGTM ! 
(I guess we should open an issue to fix it at some point ?)

PR #325, opened 2025-05-26
Remove usage of user journey in suggestion after first message.

### LINKED ISSUE #319: Do not consider click history when there are already messages in the thread
Requested by the product owner

WonderPG:
Closes #319 

I had a discussion with Fabien, he prefers if this endpoint remains one. I do think that we should eventually split them back in two endpoints since both purposes diverge more and more.

Now the use-cases are as follow:
| thread_id/click_history        | provided                             | not provided                   |
|--------------------------------|--------------------------------------|--------------------------------|
| provided and contains messages | in chat = useless history is ignored | in chat = message history used |
| provided but empty             | not in chat click history used       | error                          |
| not provided                   | not in chat click history used       | error                          |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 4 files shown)


jankrepl (approved):
LGTM:)

PR #347, opened 2025-06-10
Add thumbnail generation tool

### LINKED ISSUE #343: (Re)implement plot morphology tool
It was deleted here #330 with the idea that the bulk of the logic would be implemented in OBI one - see https://github.com/openbraininstitute/obi-one/issues/135

Image

Attention, we might cancel this ticket in favor of #344. However, the thumbnail generator service does not allow for much customization. Please make an educated decision.

### LINKED ISSUE #344: Create tools using thumbnail generator service
Please use the same logic as with other services - (e.g. autogen,  one tool per endpoint, ...)

https://staging.openbraininstitute.org/api/thumbnail-generation/docs

Note that those endpoints return raw bytes (`content-type: image/png`) so IMO what our wrapper tool should do is to fetch the image, put it on the storage and then return the storage id. This way our plotting mechanism will work perfectly.

WonderPG:
Closes #343 closes #344


jankrepl (approved):
LGTM:)

PR #360, opened 2025-06-12
Speedup ci

### LINKED ISSUE #357: Switch to uv (and make CI faster)
https://docs.astral.sh/uv/guides/integration/github/

Note that we can do it both for linting and unit tests/mypy

See below a time distrubution of unit/tests mypy pipeline

Image

### LINKED ISSUE #358: Run pytest and mypy in parallel in the CI
Should speed things up.

WonderPG:
Closes #357 closes #358 

Introducing `uv`: 48s -> 9s (down to 2s when hitting the cache)
parallelizing mypy and tests: 1m50 -> mypy 52s, unit-tests 59s so 1m to do both since they run in parallel.

Overall it went from ~2m10 to 1m20 since the `CI Frontend` is now the longest running job and takes 1m20.


.github/workflows/ci.yaml
```diff
@@ -44,27 +44,60 @@ jobs:
           bandit -qr -c pyproject.toml src/
           ruff check src/ tests/
           ruff format --check src/ tests/
```
jankrepl, line 46: Just to be consistent, can we also use `uv` for the `linting` job?

jankrepl (approved):
LGTM:) Thank you! @WonderPG

PR #365, opened 2025-06-16
migrate me-models endpoints to entitycore

### LINKED ISSUE #361: ME model tools from entitycore (instead of BlueNaas)
**What?**
Add the following
```
backend/src/neuroagent/tools/entitycore_memodel_getall.py
backend/src/neuroagent/tools/entitycore_memodel_getone.py
```

Please follow the same naming/logic conventions as for all the other `entitycore` tools

and remove the following

```
backend/src/neuroagent/tools/bluenaas_memodel_getall.py
backend/src/neuroagent/tools/bluenaas_memodel_getone.py
```

**Why only now?**
I just checked and staging entitycore does seem to have `memodel` populated now. Also, the `memodel` endpoints will be removed from bluenaas.

BoBer78:
Closes #361

### DIFF
--- a/backend/src/neuroagent/tools/entitycore_memodel_getall.py
+++ b/backend/src/neuroagent/tools/entitycore_memodel_getall.py
@@ -60,7 +60,7 @@ class MEModelGetAllTool(BaseTool):
     These are models that are ready to be simulated. You will need the ID of the models to run the subsequent simulations."""
     description_frontend: ClassVar[
         str
-    ] = """Searches for Morphological Electric models (ME-Model) in the entitycore database. Use this tool ewhen you want to :
+    ] = """Searches for Morphological Electric models (ME-Model) in the entitycore database. Use this tool when you want to:
     • Find MeModels in specific brain regions
     • Filter by subject and species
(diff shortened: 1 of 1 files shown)


backend/src/neuroagent/tools/entitycore_memodel_getall.py
```diff
@@ -0,0 +1,116 @@
...
+from neuroagent.tools.autogenerated_types.entitycore import (
+    ListResponseMEModelRead,
+    ReadManyMemodelGetParametersQuery,
+)
+from neuroagent.tools.base_tool import BaseTool, EntitycoreMetadata
+
+logger = logging.getLogger(__name__)
+
+
+class MeModelGetAllInput(ReadManyMemodelGetParametersQuery):
```
jankrepl, line 19: I would call it `ME...` instead of `Me` - to be consistent with the meaning + entitycore class naming

jankrepl (approved):
LGTM:) I would definitely merge it since the entitycore will be the source of truth from now on.

Thank you @BoBer78

WonderPG (approved):

PR #366, opened 2025-06-16
Unique tc

### LINKED ISSUE #363: Create our own primary key for the tool calls
We still need to store the `tool_call_id` returned by the provider because the `TOOL` message has to be associated to its `AI_TOOL`.

WonderPG:
Closes #363


WonderPG:
Handled in another PR

PR #371, opened 2025-06-18
Switch to using UUID where possible

### LINKED ISSUE #369: Consider using UUID as primary key type instead of VARCHAR
Please read up on the topic. Also, maybe it is a bad idea since the migration script would invalidate all the data.

* official docs on the UUID type in postgres: https://www.postgresql.org/docs/current/datatype-uuid.html
* entitycore uses UUID: https://github.com/openbraininstitute/entitycore/blob/main/app/db/model.py

I think one of the reasons why we did not use the UUID type from the very beginning was that we were trying to support both SQLite and Postgres.

WonderPG:
Closes #369


backend/alembic/versions/529e44b33a67_switch_to_uuid.py
```diff
@@ -0,0 +1,260 @@
+"""Switch to uuid
+
+Revision ID: 529e44b33a67
+Revises: 0ea274d7a584
```
jankrepl, line 4: Probably needs to be changed here too?

Not exactly sure whether this one is used by alembic but just to be sure

WonderPG: Done in [54e507d](https://github.com/openbraininstitute/neuroagent/pull/371/commits/54e507db72fc2d1a47b42ac121969e416a43070a) as well.

jankrepl (approved):
LGTM:) Thank you!

BoBer78 (approved):
LGTM !

BoBer78 (approved):

PR #377, opened 2025-06-23
Ephys tool from obi-one

### LINKED ISSUE #375: Tool for `electrophysiologyrecording-metrics/{trace_id}` from obi-one
The endpoint was added here: https://github.com/openbraininstitute/obi-one/pull/180

Note that currently the only valid input ids are those of `electrical-cell-recording` and one **cannot** give it simulated traces.

Also, I think we cannot assume the model will magically understand that `electrical-cell-recording` and `trace` refer to the same thing so we should probably write it in the description.

jankrepl:
Closes #375 
# For reviewers
* The endpoint will give you an empty result for human traces - the obi-one maintainers said they would make error messages nicer in the next iteration. So it is better to only filter for traces that are not human. Note that I stumbled upon an issue in entitycore when trying to run that query (https://github.com/openbraininstitute/entitycore/issues/250) so maybe the easiest way to test this PR is to just use this trace id `f3948959-9d55-4fef-840d-b6b7e36e92e1`
* I did some renaming - feel free to protest

# TODO
- [x] Fixed  - As soon as I ran `datamodel-codegen` I detected a problem in obi-one (https://github.com/openbraininstitute/obi-one/issues/249) that breaks our morphometrics tool
- [x] Provide a real example of response body in the unit test (right now it is just completely made up - not that it matteres). This is a blocker since you tried to find a trace id that would make the endpoint work but failed to do so - Reported one of the issues here: https://github.com/openbraininstitute/obi-one/issues/253


WonderPG (commented):
Looks good to me, I will play a bit with it still but I don't see what can go wrong on our side.
One thing I dislike about this new endpoint in obi-one is that it does the exact same thing as our code did before, in fact it is our old code. Some things were not good in this old code imo, for instance defaulting to every single feature and every single stimulus if nothing is provided.

Also I expect some changes to come since imo they messed up by not exposing the actual parameters in the API. I suggest we keep this one open for a bit since changes will need to occur once they fix it.

backend/src/neuroagent/tools/obione_ephysmetrics_getone.py
```diff
@@ -0,0 +1,85 @@
...
+        headers: dict[str, str] = {}
+        if self.metadata.vlab_id is not None:
+            headers["virtual_lab_id"] = self.metadata.vlab_id
+        if self.metadata.project_id is not None:
+            headers["project_id"] = self.metadata.project_id
+
+        ephys_metrics_response = await self.metadata.httpx_client.get(
+            url=f"{self.metadata.obi_one_url}/declared/electrophysiologyrecording-metrics/{self.input_schema.electrical_cell_recording_id}",
+            headers=headers,
+        )
```
BoBer78, line 71: Actually, Obi-one support all previous parameters we had before no ?

jankrepl: Oh shoot. I completely forgot to insert the query params. Great find!

PR #378, opened 2025-06-24
Reasoning

### LINKED ISSUE #372: Handle reasoning in the frontend
`useChat` has an option to handle reasoning I believe. It would be cool if it could be displayed in the frontend when we use a reasoning model. Ideally it shouldn't take too much space.

WonderPG:
Closes #372 

# How to test
- Select the model `Google: Gemini 2.5 flash preview (05-20) (thinking)`
- Use this query "Find morphologies in the thalamus, plot them and find morphometrics on them" It should have a mix of reasoning + tool calls + text interlaced.


BoBer78 (approved):
LGTM 🦫

PR #381, opened 2025-06-25
Add rate limit endpoint

### LINKED ISSUE #376: `GET /rate-limits` endpoint
The idea of it would be to return a JSON containing all rate limit information we have for that given user.

Note that currently we only send information about rate limits in response headers of `qa/chat_streamed` and `qa/question_suggestions`. However, you can only see how requests you have left by actually using one request.

WonderPG:
Closes #376


jankrepl (approved):
Works great:) Thank you very much! @WonderPG

PR #390, opened 2025-06-30
Only allow within brain region entitycore queries

### LINKED ISSUE #389: Force the LLM to use `within_brain_region_id` when applicable.
The frontend always uses the `within_brain_region_id` parameter when sending an entitycore request. We should make sure that when our LLM has a BR id, it passes it to the associated `within_` argument. In applicable tools feel free to use the following syntax:
```
brain_region__id: SkipJsonSchema[None] = Field(None, exclude=True)
```
to remove all of the parameters we don't want the LLM to use.
Please check if only the brain regions have this `within` mechanism, else do it where it is applicable.

### LINKED ISSUE #392: `GET tools/{name}` should honor `SkipJsonSchema`
The #389 lead to us using the `SkipJsonSchema`, however, the way the `GET tools/{name}` endpoint works is that considers all `.model_fields` 

https://github.com/openbraininstitute/neuroagent/blob/35e04c24e59f7e6125557c5b3f3b3178ffa0b2b5/backend/src/neuroagent/app/routers/tools.py#L161-L163

Please exclude those fields that have the skip json schema

jankrepl:
Closes #389  and closes #392

For reference, I decided to go for an approach that is recommended here: https://github.com/pydantic/pydantic/discussions/6699#discussioncomment-8642547


WonderPG (approved):
🦫

BoBer78 (approved):
🦫

PR #393, opened 2025-07-01
System prompt assembled from rules

### LINKED ISSUE #350: Prevent the LLM from showing asset links since they are not clickable
Image

The links created here lead nowhere. The LLM shouldn't display them.

### LINKED ISSUE #387: Assemble system prompt dynamically using reusable rules
Very related to #320 

Motivated by https://docs.cursor.com/context/rules (must read)

# Description
The idea would be to to define reusable rules 
See below a pseudocode

```python
class Rule(BaseModel):
    name: str
    text: str
```

The final system prompt would just end up looking like

```python
system_prompt=f"{GENERAL_INFO}{'\n'.join([r.model_dump() for r in rules])"
```

# Why
* Cursor has them. Probably is a good idea.
* Give more structure to how we manage the system prompt 
* Quickly changing behavior of the agent by changing the set of rules (no need to actually do prompt engineering)

# Some other potential features
* Have an env var that will help us filter them (by name)
* Similarly to the semantic routes, we could define them in separate `json` and then dynamically load them in Python (maybe not a great idea - this is what cursor does though). But having them directly in the Python code is fine too. Note that since Cursor standardized this concept there is now a whole opensource community around these rules: https://github.com/PatrickJS/awesome-cursorrules
* We can even let the user write these rules and allow for global vs per thread rules (overkill)

jankrepl:
Closes #387  and closes #350

# LOGIC
* Each rule will be a separate `.mdc` file in the `rules` folder.
*  Note that there is also the description in the "frontmatter" that is there only for us - it won't make it to the prompt (similarly to the file name)
* Make sure each rule starts with Heading 1 (# in markdown)

# TODO
- [x] implement the feature
- [x] add unit test(s)

### DIFF
--- a/backend/src/neuroagent/app/dependencies.py
+++ b/backend/src/neuroagent/app/dependencies.py
@@ -434,11 +434,7 @@ Current time: {datetime.now(timezone.utc).isoformat()}
 
     # Check if rules directory exists
     if not rules_dir.exists():
-        logger.warning(f"Rules directory not found: {rules_dir}")
-        return (
-            system_prompt
-            + "\n## NO RULES FOUND\nNo specific rules files were found in the rules directory."
-        )
+        return system_prompt
 
     # Find all .mdc files in the rules directory
     mdc_files = list(rules_dir.glob("*.mdc"))
--- a/backend/src/neuroagent/app/dependencies.py
+++ b/backend/src/neuroagent/app/dependencies.py
@@ -4,6 +4,7 @@ import logging
 import re
 from datetime import datetime, timezone
 from functools import cache
+from pathlib import Path
 from typing import Annotated, Any, AsyncIterator
 
 import boto3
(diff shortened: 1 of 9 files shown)


backend/src/neuroagent/app/dependencies.py
```diff
@@ -408,48 +409,74 @@ async def get_selected_tools(
         return selected_tools
 
 
+@cache
+def get_rules_dir() -> Path:
+    """Get the path to the rules directory."""
+    current_file = Path(__file__)
+    rules_dir = current_file.parent.parent / "rules"
+    return rules_dir
```
WonderPG, line 417: With the dependencies ever expanding, is it needed to make this a dependency ? Wouldn't it be simpler in `get_system_prompt` to do:
```
rules_dir = Path(__file__).parent.parent / "rules"
```
?

backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -0,0 +1,9 @@
+---
+description: Entitycore tools
+---
+# EntityCore Tool Guidelines
+
+## EntityCore Tool Parameters
+- For entitycore tools, make sure to use the `within_brain_region_brain_region_id` parameter if the user asks for a specific brain region
+- The `brain_region__id` parameter is **not valid** and should not be used
+- To find a brain region id of a given brain region you can use the `entity-brainregion-getall` tool and then the `name__ilike` parameter of that tool
```
WonderPG, line 9: Could we add here that brain region names must always be capitalized ? E.g. if `thalamus` is used in field `brain_region_name`, entitycore fails but `Thalamus` works.

backend/src/neuroagent/rules/tool-calling.mdc
```diff
@@ -0,0 +1,15 @@
...
+# Tool Calling Rules
+
+## Information Sources
+- **ONLY use information from tool outputs** - do not add external knowledge
+- Use only tool outputs as authoritative sources
+
+## Link and Reference Policies
+- **NEVER generate fake links** or links to storage IDs
+- **NEVER reference storage IDs** or create placeholder links
+- Only reference actual, valid URLs and resources
```
WonderPG, line 13: Maybe adding a line here would solve #350 ?

backend/src/neuroagent/rules/entitycore-tools.mdc
WonderPG, line 9: I disagree with the use of `name__ilike`. This is useful only when a partial match is required. If we provide `thalamus` to it it will find a bunch of subregions (sorted alphabetically) but not the one we expect. Imo people will more often than not use the exact name of the BR they are after. This is valid at least for the `brainregion-getall` tool, maybe not for the others.

WonderPG (commented):
Thanks for the PR ! Two small things:
- I heard many time that it is important to re-affirm essential elements at the end of a system prompt to "remind" the LLM. is this handled in any way in cursor ? Should we handle it in any way ?
- Could you write a short README in the `rules` folder to describe in a few lines how to write a new rule to fit the format your code expect ?
Thank you !

backend/src/neuroagent/app/dependencies.py
WonderPG: Ah I see. It's fine then :)

jankrepl:
Readme was added here: 6b052fd1

backend/src/neuroagent/rules/entitycore-tools.mdc
jankrepl: Well so I instructed it to use `name__ilike`. Here I added a new commit that also mentions the `name` and the need to capitalize

861e0649

WonderPG (dismissed):

PR #394, opened 2025-07-01
Auto upgrade alembic

### LINKED ISSUE #289: Investigate the correct way to `alembic upgrade head` in deployments
Currently we need to tunnel the rds db and apply locally `alembic upgrade head` when we deploy a version that changed the DB schema. This causes downtime in staging and prod. I asked around and they seem to automatically run when the docker image is being deployed.

An example from the vlab api, they have it in the `docker-entrypoint.sh`: https://github.com/openbraininstitute/virtual-lab-api/blob/main/docker-entrypoint.sh

If there is a better way you can think of, feel free to document it/apply it here.

WonderPG:
Closes #289 

Inspired from 
https://github.com/openbraininstitute/virtual-lab-api/blob/main/Dockerfile
https://github.com/openbraininstitute/virtual-lab-api/blob/main/docker-entrypoint.sh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reasoning component.
 - Rate limit endpoint.
 - More entity core endpoints.
+- Auto alembic upgrade.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


BoBer78 (approved):
LGTM !

jankrepl (dismissed):

PR #401, opened 2025-07-02
Make tool selection persist in between conversations

WonderPG:


BoBer78 (dismissed):
It works fine, do we want to have this behavior of the tool being kept between conversations ?

BoBer78 (approved):

PR #405, opened 2025-07-02
Try to emphasize distinction between entitycore and LS

### LINKED ISSUE #399: Entitycore vs web/literature search - make the distinction clear to the LLM
The LLM sometimes overestimates the power of entitycore and thinks that it contains things it does not contain (e.g. for a given brain region what functions is it responsible for - like walking, talking,...).

Where to make the change? I don't know. Multiple options

* Inside the `entitycore-tools.mdc`
* Inside entitycore tool description

WonderPG:
Closes #399 

I could not reproduce the behavior the user got. It's a first iteration we will see if this rule is enough.

The worst I manage to get was that the LLM called an entitycore tool when not needed, but immediately after called literature
image


jankrepl (approved):
Great! Than you @WonderPG

PR #406, opened 2025-07-02
Navigation links

BoBer78:
More of a test, does work fine with `gpt-4.1`. Smaller models struggle to really understand the details of the system prompt.

### DIFF
--- a/backend/src/neuroagent/rules/entitycore-tools.mdc
+++ b/backend/src/neuroagent/rules/entitycore-tools.mdc
@@ -8,6 +8,34 @@ description: Entitycore tools
 - The `brain_region__id` parameter is **not valid** and should not be used
 - To find a brain region id of a given brain region you can use the `entity-brainregion-getall` and provide the `name` parameter but make
 sure to capitalize the first letter of the brain region name.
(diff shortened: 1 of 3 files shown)


backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -8,6 +8,34 @@ description: Entitycore tools
...
+   - `morphology` - for reconstruction morphology data
+   - `electrophysiology` - for electrical traces
+   - `neuron-density` - for density measurements
+   - Other asset types as applicable
+
+3. **Entity ID:**
+   Conclude with the specific asset ID
+
+**Example Link:**
+- `https://staging.openbraininstitute.org/app/virtual-lab/explore/interactive/experimental/morphology/afa2434f-1ccc-4fff-a047-d399e59768c9`
```
jankrepl, line 37: I know this is a prototype. But the big problem here is that the frontend url will need to be injected via an env var. 

Currently, we don't have a mechanism that would allow us to inject env vars into the system prompt. We can introduce it (or we go for the tool approach)

PR #409, opened 2025-07-03
Bot autogen

### LINKED ISSUE #400: Autogen bot PR
It would be nice if everytime there is a change the a 3rd party API the autogen would be rerun automatically and a PR created with all the changes.

Then a developer can approve or jump in and extra logic.

It is related to #396 that would only detect that there is a problem - but not attempt to fix it.

Why is it better then doing this manually?
I am just trying to save us time. Like in theory we would have to create a ticket for every new autogen and then discuss who does it.

jankrepl:
Closes #400


jankrepl:
Closing this. Since the current setup is fine.

PR #410, opened 2025-07-03
Add python interpreter

### LINKED ISSUE #408: Python interpreter tool (or just a calculator)
# Why
The main usecase for now is just a very powerful calculator. I would say we are not adding this to run OBI related code just yet (since generating OBI code is a ticket on its own #172). This is purely about executing a piece of code (in most cases even without 3rd party Python packages).

# How
Some suggestions were already mentioned in #342 

I think https://ai.pydantic.dev/mcp/run-python/ is really the way to go (and since it is an MCP server adding it is a breeze)

Please double check that the MCP server somehow handles the below things
* Each tool call happens in a completely separate environment (e.g. if user Alice saves a file in `/tmp/whatever` then another user Bob cannot access the file)
* Maybe allowing all python functionality is not what we want (e.g. `subprocess`, `os` module) - but I guess you just study all the flags the MCP server has and maybe there are some useful ones. The obvious "attacks" one can do is
   * Read all the environment variables in the running container
   * Remove all the files in the filesystem
   * Kill processes (e.g. the backend process itself)
   * Infinite recursion/loops - like there should be some timeout logic
* Network egress? Should it be allowed? Like should we allow the tool to send out requests to random APIs

# Comment
* If you think having a Python interpreter is an overkill (or it simply is not worth all the potential issues) I am fine if we just use some calculator mcp server

WonderPG:
Closes #408 

I went with the python interpreter because it is very powerful from my experiments and the calculator cannot go beyond basic operations. E.g. you can't compute a square root that easily. Imagine more complex stuff. The python interpreter can use external libraries easily.

I tried to breach security to test, the LLM refused calls that asked to remove files from the computer. Another issue is the possibility to generate very large outputs. E.g. "Loop for one million iterations in python and print the value of the iterator each time" is something the LLM would do. Since we are not charging the user per token yet, I wrote a rule to restrict that. The LLM seems to follow it but it might be possible to trick it.

## How to test
```bash
brew install deno
```


backend/Dockerfile
```diff
@@ -10,17 +10,26 @@ RUN apt-get -y install curl \
...
+
+# Create package.json and install pyodide locally
+WORKDIR /code
+RUN echo '{ "type": "module" }' > package.json
+RUN npm install pyodide
+RUN cp -r node_modules /node_modules
+RUN mv setup_pyodide.js /setup_pyodide.js && mv package.json /package.json
+WORKDIR /
 RUN chmod +x /docker-entrypoint.sh
+RUN chmod +x /setup_pyodide.js
```
jankrepl, line 30: Don't think you need to add any extra permissions since you are running it via `node`

backend/Dockerfile
```diff
@@ -10,17 +10,26 @@ RUN apt-get -y install curl \
...
 
 RUN mv /code/alembic /alembic
 RUN mv /code/alembic.ini /alembic.ini
 RUN mv /code/docker-entrypoint.sh /docker-entrypoint.sh
+
+# Create package.json and install pyodide locally
+WORKDIR /code
+RUN echo '{ "type": "module" }' > package.json
+RUN npm install pyodide
+RUN cp -r node_modules /node_modules
```
jankrepl, line 26: why not `mv` here too?

backend/Dockerfile
WonderPG: Done in [be9465d](https://github.com/openbraininstitute/neuroagent/pull/410/commits/be9465d51646887cc1e19e58a51fbfd9e1c9fce3)

backend/Dockerfile
WonderPG: Done in [be9465d](https://github.com/openbraininstitute/neuroagent/pull/410/commits/be9465d51646887cc1e19e58a51fbfd9e1c9fce3)

docker-compose.yml
```diff
@@ -6,7 +6,12 @@ services:
     ports:
       - "8078:8078"
     environment:
-      - NEUROAGENT_DB__PREFIX=postgresql+asyncpg://postgres:pwd@postgres:5432/neuroagent
+      - NEUROAGENT_DB__PREFIX=postgresql+asyncpg://
+      - NEUROAGENT_DB__HOST=localhost
```
jankrepl, line 10: This should be `postgres` since we use the internal docker compose network. (I got an error without it but not sure why - localhost should work too)...anyway

docker-compose.yml
```diff
@@ -6,7 +6,12 @@ services:
...
+      - NEUROAGENT_DB__PORT=5432
+      - NEUROAGENT_DB__USER=postgres
+      - NEUROAGENT_DB__PASSWORD=pwd
+      - NEUROAGENT_DB__NAME=neuroagent
       - NEUROAGENT_KEYCLOAK__ISSUER=${KEYCLOAK_ISSUER}
       - NEUROAGENT_TOOLS__LITERATURE__URL=https://whatever.com
       - NEUROAGENT_OPENAI__TOKEN=${OPENAI_API_KEY}
```
jankrepl, line 17: I know this is unrelated...but this env variable is not named like that anymore.


It should be `NEUROAGENT_LLM__OPENAI_TOKEN`

backend/setup_pyodide.mjs
```diff
@@ -0,0 +1,17 @@
...
+import { loadPyodide } from "pyodide";
+async function run() {
+    const pyodide = await loadPyodide();
+
+    await pyodide.loadPackage("micropip");
+
+    const micropip = pyodide.pyimport("micropip");
+
+    await micropip.install([
+    "typing-extensions",
```
jankrepl, line 11: You can remove it...it is a dependency of pydantic so it will be installed automatically https://github.com/pydantic/pydantic/blob/dac3c437870a056839c0a485ae3112c76d5e4a7f/pyproject.toml#L47

backend/src/neuroagent/mcp.json
```diff
@@ -15,5 +15,9 @@
       "env": {
         "EXA_API_KEY": "NEUROAGENT_MCP__SECRETS__EXA_API_KEY"
       }
+    },
+    "run_python":{
+      "command": "deno",
+      "args": ["run", "-R=node_modules", "--node-modules-dir=auto", "jsr:@pydantic/mcp-run-python", "stdio"]
```
jankrepl, line 21: So as we need, when the server is launched it will create `node_modules/` in the folder you are launching it in. I suggest we add `node_modules` to the `backend/.gitignore` (we first need to create it). Mostly to avoid accidental tracking with git

docker-compose.yml
WonderPG: I commited the wrong version apparently, I went one cmd+z too far. Good catch !

backend/src/neuroagent/mcp.json
WonderPG: Done in 983d04c.

jankrepl (approved):
Works great:) Thank you @WonderPG

BoBer78 (approved):

PR #411, opened 2025-07-04
Update false positive utterances

jankrepl:
based on #278


backend/src/neuroagent/fixed_llm_responses.yml
```diff
@@ -38,11 +38,20 @@ routes:
       - "Can you get me 2 electrial cell recordings?"
       - "Can you get me more ?"
       - "Can you help me find some morphologies in the platform?"
+      - "can you only look for morphologies in this brain region?062ebc92-cebb-498a-a76d-b5e1bab02cf"
```
WonderPG, line 41: Even though the message had the uuid, I would maybe remove it here. The utterance is stronger without imo. What do you think ?

WonderPG (commented):
LGTM !

backend/src/neuroagent/fixed_llm_responses.yml
jankrepl: Well the funny thing is that without it the human route is not triggered. So it needs to be there.

WonderPG (approved):

PR #412, opened 2025-07-04
General guidance for entitycore tools

### LINKED ISSUE #404: Entitycore rule  - explain general principles
Let's help the LLM and mention the following things that are shared across all entitycore endpoints

* Pagination mechanism in the GET ALL endpoints and ask it to ask to user whether they want to fetch more data (and potentially mention the number of results that is found in the response body of every page)
* GET ALL endpoints won't have assets unlike their GET ONE counterparts. I know we do it on the tool description level but it is becoming a standard so we might as well pull it out to a rule

Probably related to #399

WonderPG:
Closes #404


backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -8,10 +8,16 @@ description: Entitycore tools
...
+- To find a brain region ID: use `entitycore-brainregion-getall` with the `name` parameter (capitalize the first letter of the brain region name).
+- **MANDATORY**: Always resolve brain region names to their unique IDs before any data queries.
+- **CRITICAL RULE**: Once you have a brain region ID, you MUST use `within_brain_region_brain_region_id` with the UUID for ALL subsequent queries.
+- **STRICTLY FORBIDDEN**: Using `brain_region__name` or any name-based parameters when an ID is available.
+- **NEVER** fall back to name searches when you already have the ID in memory.
+- **INVALID PARAMETER**: `brain_region__id` does not exist and must not be used.
+
+## GET ALL Endpoint Guidelines
+- When calling `GET ALL` endpoints, please mention the `total_items` count and ask the user if they want to see additional data before using the `pagination` parameter to retrieve subsequent pages.
+- If a `GET ALL` endpoint returns an empty asset list, call the corresponding `GET ONE` endpoint or `GET ALL Assets` to retrieve the actual assets for that entity.
```
jankrepl, line 20: Will this not confuse the model? Like the idea is for the model to we aware of it but not necessarily to follow up GET ALL with a bunch of GET ONE just to get the assets without even the user asking for it

WonderPG: I changed it to `- If a `GET ALL` endpoint returns an empty asset list, they can be accessed through the corresponding `GET ONE` endpoint or `GET ALL Assets` if needed.`. Let me know if you prefer it

BoBer78 (approved):

PR #423, opened 2025-07-10
Allow to override tool name/description

### LINKED ISSUE #415: Enable custom tool names (+ descriptions) for MCP servers
Right now we basically just take `{server_name}-{tool_name}` with some cosmetic changes. But maybe we should make it more flexible and let the user specify (optionally) custom names for each of the tools in `mcp.json`.

Bonus: Allow for custom tool descriptions (in case the official ones are bad). I would not allow for parameter definitions. Just the general tool description.

WonderPG:
Closes #415 

It seems like the official MCP library doesn't support any kind of naming override. I had to hack it a bit which is not the cleanest but I guess it works

## How it works
- The `name` and `name_frontend` are only made out of the `tool_name` (i.e. the `server_name` is not used anymore)
- One can define tool name/description override directly in `mcp.json` as follow:
```json
{
      "my_server" :{
            "command": "my_command",
            "args": ["my", "list", "of", "args"],
            "env": {
                    "MY_ENV_KEY": "MY_ENV_VALUE"
                  },
            "tool_metadata": {
              "tool_name_1": {
                "name": "my-custom-name-for-tool-1",
                "name_frontend": "Great Name",
                "description": "This is my own description for tool 1 that the LLM sees."
                },
             "tool_name_2": {
                "name": "my-custom-name-for-tool-2",
              },
            "tool_name_3": {
                "description": "This is my own description for tool 3 that the LLM sees.",
                "name_frontend": "My Tool 3"
                }
          }
```
The keys `tool_name_1`, `tool_name_2` etc... must be the original name of the tools, as provided by the server. To find the original tool name, go to `/tools` in our frontend. The `tool slug` is the original name.
The key `tool_metadata` is optional.
Each of the keys `name`, 1name_frontend` and `description` are also optional. No need to have both every time. The omitted ones will take the original value.

I added some examples in `mcp.json`. These are just examples to demonstrate how the PR works, they are not meant to be kept as is. Please feel free to come up with names/descriptions of tools you would like to change, either here or in another PR.


backend/src/neuroagent/mcp.py
```diff
@@ -123,12 +155,9 @@ class Metadata(BaseMetadata):
...
     # Create the tool class
     class MCPDynamicTool(BaseTool):
-        name: ClassVar[str] = f"mcp-{server_name.replace(' ', '-')}-{tool_name}"
+        name: ClassVar[str] = tool_name
         name_frontend: ClassVar[str] = " ".join(
-            [
-                word.capitalize()
-                for word in re.split(r"[-/_=+*]", f"{server_name}-{tool_name}")
-            ]
+            [word.capitalize() for word in re.split(r"[-/_=+*]", tool_name)]
```
jankrepl, line 160: IMO we should give the user the chance to specify the `name_frontend` as well in `mcp.json`. 

This "beautify" logic is random IMO

WonderPG: Done in e564aa2

BoBer78 (dismissed):
I tested it briefly and it works fine

jankrepl (approved):
LGTM:) Thank you @WonderPG

PR #430, opened 2025-07-14
Optionally allow for exclusion of empty threads

### LINKED ISSUE #425: Exclude empty threads (optionally)
Introduce a new query parameter `GET /threads/exclude_empty=true` (by default `false`). If set to `true` we modify the SQL query to only yield those threads that have at least 1 message associated to them (it will require a join).

WonderPG:
Closes #425


jankrepl (approved):
LGTM:)

PR #437, opened 2025-07-16
Don't send boto3 HTTP requests in tests 

jankrepl:
# Description
When we instantiate `boto3.client("s3")`  (done in our `get_s3_client` dependency) a request is automatically sent to AWS. One can disable it via the `AWS_EC2_METADATA_DISABLED` env var https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html

* Note that since this request was not made by `httpx` and that is why `httpx_mock` never caught it
* Why did we trigger the `get_s3_client` dependency in some tests? Well, it is used in context variables. Which is used in the `qa/chat_streamed` endpoint which is tested here `tests/app/routers/test_qa.py::test_chat_streamed` (+ some other places).


BoBer78 (approved):
Works great !

PR #438, opened 2025-07-16
Creation date filter in GET /threads

### LINKED ISSUE #436: Filter threads based on creation date (optionally)
Make sure to use the best practices and probably even copy what entitycore does to make sure we have the same logic across all microservices

jankrepl:
Closes #436 

When working on this I realized we need #439 (to be addressed in a separate ticket)

### DIFF
--- a/backend/tests/app/routers/test_threads.py
+++ b/backend/tests/app/routers/test_threads.py
@@ -705,6 +705,14 @@ def test_get_threads_creation_date_filters(
             response = raw_response.json()
             thread_responses.append(response)
 
+        # TODO: This behavior is temporary. In the future, the API should return timezone-aware
+        # creation dates (with UTC timezone). When that change is made, these assertions should be
+        # updated to expect tzinfo == timezone.utc directly from the response.
+        # Verify all results have naive datetime (no timezone info)
+        for resp in thread_responses:
+            date = datetime.fromisoformat(resp["creation_date"])
+            assert date.tzinfo is None  # Response dates should be naive
+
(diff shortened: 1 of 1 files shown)


backend/src/neuroagent/app/routers/threads.py
```diff
@@ -144,6 +145,8 @@ async def get_threads(
     virtual_lab_id: UUID | None = None,
     project_id: UUID | None = None,
     exclude_empty: bool = False,
+    creation_date_lte: AwareDatetime | None = None,
+    creation_date_gte: AwareDatetime | None = None,
```
WonderPG, line 149: Could you elaborate on the choice of `AwareDatetime` here ? I know the original intent is to have the timezone information available, but querying these manually becomes quite hard due to the format. In LS, we used to have filters on date that had the `YYY-MM-DD` format https://github.com/openbraininstitute/scholarag/blob/da01fd636617f78e2155c472cf9856522418f6a9/src/scholarag/app/dependencies.py#L190-L203 which made it a bit simpler to query.  

Now the minimal format the user must respect is `YYYY-MM-DDT00:00:00Z` and I would argue filtering on anything more fine grained than days is not necessarily useful. I don't know if the frontend plans to expose a search bar for threads that would filter on date, but if they do so I doubt they will allow to chose up to the second. I know that the `Datetime` type of postgres goes to the sub second resolution, but the query could be done on a less fine grained basis. Just an idea.

Also I would definitely do the same as in LS here, and add something along the lines of 
```python
Query(description="Format: YYYY-MM-DDT...")
```
because a new user of our API will most likely not guess it.

Issue #379, opened 2025-06-25
Remove Literature Search tool


jankrepl:
# Description

~Please get a final confirmation from the leadership before you do so and understand what the official replacement for it is (exa's `research_paper_search` I guess). Note that there is a ticket for it here: https://github.com/openbraininstitute/prod-explore-functionality/issues/201~

UPDATE: It has been approved

# Acceptance criteria
- [x] Setup account + get API key
- [x] Remove  all the infrastructure used to deploy it - https://github.com/openbraininstitute/aws-terraform-deployment/
   - SQS, Redis, Eventbridge, S3 bucket, ECS task(s), ...
- [x] Remove card details from Cohere
- [ ] Remove card details from Tavily
- [x] Remove Tavily MCP + tool from this repo
- [x] Remove the LS tool in this repo
- [x] Rename EXA tools accordingly.

evakennyobi:
Thurs meeting to be discussed

PR #443, opened 2025-07-17
Cleanup any LS or WS related code

### LINKED ISSUE #442: Override default "Run Python" tool description
Image

The part about dependencies is not true since we disabled dynamic installation of Python dependencies. So I guess you can just keep the first part.

WonderPG:
Contributes to #379 Closes #442


jankrepl (approved):
LGTM:) Thank you!

PR #240, opened 2025-03-27
Add platform description

BoBer78:
closes #238

PR #445, opened 2025-07-21
Include Sanity data

### LINKED ISSUE #304: Video tutorial transcripts -> semantic routes

jankrepl:
Closes #240 (this is highly arguable since there will probably be other places that describe the platform somehow) and closes #304 

# Description for reviewers
* Before you start reviewing you can just check the official Sanity website. See below some specific pages that I found relatively useful
  * https://www.sanity.io/docs/http-reference/query
  * https://www.sanity.io/docs/content-lake/how-queries-work
  * https://www.sanity.io/docs/content-lake/datasets
  * https://www.sanity.io/docs/content-lake/query-cheat-sheet
* The complexity of this PR comes mostly from the fact that Sanity stores different types of documents (`_type`) and every single one of them has a different schema. I therefore introduced a `pydantic.BaseModel` for each Sanity `_type`.
* Additionally, not all Sanity attributes are useful. The ones that are useful unfortunately don't follow some fixed case conventions and therefore I decided to introduce a `sanity_mapping: dict[str, str]` that maps from neuroagent names to Sanity names.
* Additionally, some schemas contain values that are portable text: https://www.sanity.io/guides/beginners-guide-to-portable-text. To avoid giving the raw portable text to the LLM (inside tool output) I defined `flatten_portable_text` function that effectively does preorder traversal and only extracts the text. ~To trigger the flattening one needs to mark include the given field in `portable_text_attributes: list[str]`~. We use the `field_validator` mechanism to handle the full conversion - it might be more than just portable text flattening.
* Note that there are more `_type` then the ones I added. We can easily add new ones in the future.

# TODO
- [x] Add unit tests for every type (focus on field validator) - use actual sanity documents (they are public anyway)
- [x] Make sure the tool selection LLM is better at picking this tool (extra explaantion in the system prompt)
- [x] Write parsing logic
- [x] Include relevant `_type`
  - [x] `futureFeaturesItem`
  - [x] `glossaryItem`
  - [x] `news`
  - [x] `publicProjects`
  - [x] `tutorial`
  - [x] `pages`
- [x] Why is `result.model_dump()` not serializing everything? This issue happens in the actual tool output too!!!
- [x] Explain clearly in the comments what `sanity_mapping` and `portable_text_attributes` are
- [x] Make tool description amazing
- [x] Write unit tests (especially for the portable text flattening with real world example)
- [x] Fix mypy issue


WonderPG (commented):
Thanks for the PR, only minor comments !

I played a bit with it and I noticed that the tool selection phase often fails to put this tool in the mix. Imo there are 3 solutions:
- Add a paragraph in the description that describes when the tool should be used, maybe with few shot examples.
- Add a line in the system prompt of the tool selection LLM.
- Introduce "always selected" tools
- 
I would personally go for 1) then 2) if it's not enough.

Another issue I face is that the type of questions I ask to try to trigger this tool very often trigger the fixed responses.

jankrepl:
> The tool seems to work nicely !
> 
> Small comment, I got this when testing : image This is probably due to the fact that there is no tutorial for that in the platform yet.
> 
> Another small comment is that we may want to remove the platform description (or at least shorten it) from the system prompt.

I think the problem is that the LLM decided to provide "query" which does keyword matching. If it left it blank it would show all the tutorials.

backend/src/neuroagent/app/config.py
```diff
@@ -90,12 +90,27 @@ class SettingsEntityCore(BaseModel):
     model_config = ConfigDict(frozen=True)
 
 
+class SettingsSanity(BaseModel):
+    """Sanity settings."""
+
+    project_id: str = "fgi7eh1v"
+    dataset: Literal["staging", "production"] = "staging"
+    version: str = "v2025-02-19"
+    model_config = ConfigDict(frozen=True)
```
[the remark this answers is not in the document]
jankrepl: Done in b1cf0e86

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,508 @@
...
+    """
+    # Handle single block case
+    if isinstance(blocks, dict):
+        blocks = [blocks]
+
+    lines = []
+    for block in blocks:
+        # Handle blocks with children (spans)
+        if "children" in block:
+            text = "".join(child.get("text", "") for child in block.get("children", []))
```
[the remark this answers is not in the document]
jankrepl: Done here: 2b92a8ca

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,508 @@
...
+        elif "content" in block and isinstance(block["content"], (list, dict)):
+            nested_text = flatten_portable_text(block["content"])
+            if nested_text:
+                lines.append(nested_text)
+
+        # Handle arrays
+        elif isinstance(block, list):
+            nested_text = flatten_portable_text(block)
+            if nested_text:
+                lines.append(nested_text)
```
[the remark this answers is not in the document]
jankrepl: Possibly - but I don't remember seeing that overlap. Honestly, I wrote this function based on what I have seen in the actual data. So the current version of it is just my empirical plaintext serializer. I am happy to extend/modify it if we deal with a Sanity document that it does not support or if I messed up.

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,508 @@
...
+    - Sorting by creation or update date
+    - Pagination
+    - Field selection based on model's sanity_mapping
+    """
+    model = SANITY_TYPE_TO_MODEL[document_type]
+    base_query = build_base_query(document_type, query)
+
+    # Add sorting
+    sort_field = "_createdAt" if sort == "newest" else "_updatedAt"
+    sort_order = "desc" if sort == "newest" else "asc"
```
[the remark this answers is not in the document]
jankrepl: That is a bug:) Nice find! Fixed in 8748b402

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,508 @@
...
+    base_query += f" | order({sort_field} {sort_order})"
+
+    # Add pagination
+    start = (page - 1) * page_size
+    base_query += f"[{start}...{start + page_size}]"
```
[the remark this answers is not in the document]
jankrepl: To be consistent with entitycore

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,508 @@
...
+    # Get the appropriate model class based on document type
+    doc_type = doc["_type"]
+    model_class = SANITY_TYPE_TO_MODEL[doc_type]
+
+    # Process any portable text fields
+    processed_doc = doc.copy()
+    for field in model_class.portable_text_attributes:
+        field_value = doc.get(field)
+        if isinstance(field_value, (dict, list)):
+            processed_doc[field] = flatten_portable_text(field_value)
```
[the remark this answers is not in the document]
jankrepl: Addressed in a bunch of recent commits.

I completely got rid of the concept of `portable_text_attributes` and instead wrote a custom `field_validator` for all the tricky "fields" that contain nasty nested JSONs.

I also added a lot of unit tests using actually sanity data making sure these validators do the right thing.

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,666 @@
...
+
+    @field_validator("definition", mode="before")
+    @classmethod
+    def flatten_definition(cls, v: list[dict[str, Any]] | None) -> str | None:
+        """Flatten the definition from portable text blocks into a single string."""
+        if v is None:
+            return None
+
+        # Use the existing flatten_portable_text function to process the definition
+        flattened = flatten_portable_text(v)
```
WonderPG, line 150: The function `flatten_portable_text` already does the check for None
```python
if blocks is None or not isinstance(blocks, (list, dict)):
        return blocks
```

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,666 @@
...
+
+        # The flatten_portable_text function should return a string for portable text
+        if isinstance(flattened, str):
+            return flattened
+
+        # Fallback: convert to string
+        return str(flattened) if flattened else ""
```
WonderPG, line 157: In theory `flatten_portable_text` should return a string if the block is not `None` right ? Shouldn't we instead raise an error if it is not a string so that we can catch potential bugs ?

backend/src/neuroagent/tools/obi_expert.py
WonderPG: Same for the other validators

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,666 @@
...
+    description: str
+    scale: str
+
+    sanity_mapping: ClassVar[dict[str, str]] = {
+        **SanityDocument.sanity_mapping,
+        "topic": "Topic",
+        "feature_title": "Feature_title",
+        "description": "Description",
+        "scale": "Scale",
+    }
```
WonderPG, line 174: This class doesn't need the validator, `description` is always a string ?

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,666 @@
...
+                if "content" in block and isinstance(block["content"], list):
+                    for content_block in block["content"]:
+                        if (
+                            isinstance(content_block, dict)
+                            and content_block.get("_type") == "block"
+                        ):
+                            if "children" in content_block:
+                                for child in content_block["children"]:
+                                    if isinstance(child, dict) and "text" in child:
+                                        text_parts.append(child["text"])
```
WonderPG, line 292: So the type `richContent` is not a portable text ? How comes before switching to validator based flattening, we didn't have this logic that is different than the `flatten_portable_text` one ?

backend/src/neuroagent/tools/obi_expert.py
```diff
@@ -0,0 +1,666 @@
...
+    - Optional text matching across all mapped fields
+    - Sorting by creation or update date
+    - Pagination
+    - Field selection based on model's sanity_mapping
+    """
+    model = SANITY_TYPE_TO_MODEL[document_type]
+    base_query = build_base_query(document_type, query)
+
+    # Add sorting
+    sort_field = "_createdAt"
```
WonderPG, line 448: Shouldn't it be `_modifiedAt` ? Instinctively I would do it this way but with low confidence.

WonderPG (commented):
Just small questions, thanks for the refactor !

backend/src/neuroagent/tools/obi_expert.py
jankrepl: Done in a1c648ae

backend/src/neuroagent/tools/obi_expert.py
jankrepl: Yes. Exactly.

backend/src/neuroagent/tools/obi_expert.py
WonderPG: Fair.

WonderPG (dismissed):

PR #446, opened 2025-07-21
Add copy button for tool output

### LINKED ISSUE #444: Add copy to clipboard button to tool outputs
Some of the tool outputs are huge and it selecting the entire JSON is not easy. Let's add a copy to clipboard button so that people can view the JSON in their preferred editor

WonderPG:
Closes #444


BoBer78 (approved):
🦫

PR #450, opened 2025-07-28
Entitycore links

### LINKED ISSUE #346: Navigate tool
This ticket is blocked by the actual frontend completely switching to entitycore and using entitycore identifiers everywhere.

**Idea**

We introduce a new env var - frontend url.

The tool itself will be a full url generator that takes in 
* `entity_route: Literal["reconstruction-morphology", ...]`
* `entity_id: UUID`

And it spits out a valid URL in the frontend. You can figure out those urls by just browsing the frontend or checking the actual Next.js code and inspecting the page hierarchy. The user could then click on it in the chat if they want to be redirected there.

For example, see below a url of a reconstruction morphology

```python
f"{frontend_url}/app/virtual-lab/explore/interactive/experimental/morphology/{morphology_id}
```

BoBer78:
closes #346


BoBer78:
While overriding the autogen tool outputs is very annoying, this version works way better than the one utilizing a tool to generate the links.

Minor issue: the LLM struggles to make the name clickable, and adds a link to the entity in the list of features.

PR #452, opened 2025-07-29
Entitycore links (tool version)

BoBer78:


jankrepl:
We had an internal discussion and decided to go for the other approach #450

PR #463, opened 2025-08-07
Add utterance structure

### LINKED ISSUE #458: Add utterances to BaseTool to allow for few-shot prompting
# Description
The idea would be to add `utterances: ClassVar[list[str]] = []` in the `BaseTool`. And then all our tools (the children of `BaseTool`) could redefine that variable and it would contain user messages that should trigger that tool.

# Why (potentially) useful
* These utterances can be then dynamically inserted into various LLM system prompts (either the agent LLM but most likely just the tool selection LLM) and make them perform way better.
* Separation of concerts: `description` holds description of what the tool does, `utterances` hold examples of human messages that should trigger it
* We are getting rid of the semantic router -> so giving few shot examples (via utterances) to LLMs might make the transition smoother #457 

# What parts of the code are influenced?
* `BaseTool`
* Tool selection LLM
* Agent LLM - if we decide to somehow merge `description` and `utterances` into `description_with_utterances` - maybe not necessary
* (backend + frontend) GET `/tool/{tool_name}` - IMO we should include the utterances too.
* Individual tools - we can have separate tickets that actually add utterances. **Please sort alphabetically**
* `mpc.json` should somehow support addition of utterances

# Comments
* Maybe it is a bad idea. However, the real issue that motivated this ticket was that we wanted to start adding few shot examples to tool descriptions and ideally only make it visible to the filtering LLM (since the description alone did not lead to the LLM to select the tool when it should have). There is nothing wrong with putting the few shot examples in the `description` directly but this ticket can help us standardize the process and separate concerns

WonderPG:
Closes #458


jankrepl (dismissed):
LGTM:)

jankrepl (approved):
LGTM:)

PR #465, opened 2025-08-07
Remove semantic router

### LINKED ISSUE #457: Delete semantic router
The OBIExpert tool introduced in #445 might have arguably made semantic router redundant since we will be fetching latest info about the platform from some remote API.

We should probably discuss if there are any downsides before pulling the trigger

WonderPG:
Closes #457


jankrepl (approved):
LGTM:)

PR #473, opened 2025-08-13
Introduce Deepeval evaluation

### LINKED ISSUE #472: Deepeval evaluation (iteration 0)
https://github.com/confident-ai/deepeval

# TODO
* Read through the docs ( https://deepeval.com/docs/getting-started)  and understand all the relevant features the github repo has (we can do a small presentation for the rest of the colleagues)
* Setup a minimal evaluation dataset that contains ~10 past questions we have had in the past 
* Implement tests/ metrics
Some interesting metrics be metrics to use
* Som
* Tool correctness
* Argument correctness: 

# Comments
* We agreed to use `chat_streamed` to generate the responses and "unstream them" (going from `generator[str]` to `str`) and only then insert into deepeval
* We want to track the questions and the expected answers/tool calls but we don't want to store the evaluation metrics and generated answers in git

WonderPG:
Closes #472


backend/src/neuroagent/scripts/evaluate_agent.py
```diff
@@ -0,0 +1,443 @@
...
+                "arguments": {},  # Will be filled later
+            }
+
+        # Tool args (streamed incrementally)
+        elif prefix == "c":
+            tool_call_id = content.get("toolCallId")
+            delta = content.get("argsTextDelta", "")
+            if tool_call_id:
+                tool_args_buffer.setdefault(tool_call_id, "")
+                tool_args_buffer[tool_call_id] += delta
```
jankrepl, line 262: Are we sure we want to keep this? The `9` https://v4.ai-sdk.dev/docs/ai-sdk-ui/stream-protocol#tool-result-part will contain all the information. I don't see how in practice there would be no `9` and only `b` and a bunch of `c`

WonderPG: Done in ffa57298

PR #484, opened 2025-08-18
Enforce DeepEval updates in PRs

### LINKED ISSUE #478: Include test cases in `eval_dataset.csv`
To evaluate the agent we must add samples in the file `eval_dataset.csv`. Try to find use cases that historically were/are an issue for the LLM. To be merged after #473 is merged.

### LINKED ISSUE #553: "Deepeval updated" github action
The idea here is that we would to start forcing the developers to run deepeval locally with each PR.

What should be done
* Implement the "check updated" github action - if the user does not update it it is a fail
* Somehow nicely take the most recent results and display them in the CI (sure, there will be a file tracked with git that will hold the results it is just that it might be nicer to have it in the CI itself similarly to coverage)

Side comment
* The update changelog is effective right away and arguably, the developer can modify the changelog before implementing the feature. Deepeval is a bit different since we would ideally want to run it just before merging to main. I don't know.

jankrepl:
(Partly) closes #478  and closes #553

# Comments for reviewers

* I decided to ditch storing all **outputs** as a CSV and instead used a `JSON` that is easier to query with `jq` and definitely easier to read if you open it in a vanilla text editor
* I decided to ditch storing all  **inputs** as a CSV and instead I introduced a folder + file hierarchy
   
image

* There are 2 scripts that are relevant. See the below sketch
image

* After this is merged how will it influence our development flow? It will be very similar to how we handle the `CHANGELOG.md`. See below the exact steps we will have to do
  1. Create a new PR and implement all the features
  2. The `Check DeepEval` action will be failing. To make it pass run the below locally
      ```bash
      python -m neuroagent.scripts.evaluate_agent  $TOKEN 
      ```
  3. Stage and commit and push the two files that changed inside `backend/eval/output`
  4. The `Check DeepEval` should be passing and the standard output should be showing some statistics of what changed compared to `main`. Note that once there is a previous result on `main` see below an example of how it will actually look like
image

* There are a lot of questionable things in this setup
   * We assume that whoever runs the evaluation locally will do it right (e.g. have the identical set of env vars as all the other people on the team). IMO the most important thing is to have all tools whitelisted and point at staging microservices
   * Runnning the evaluation twice gives you different results so distinguishing between "random hickup" vs actual change in performance might not be trivial. But I guess the more test cases we add the more reliable it will get.
   * Not super sure about storing the `detail.json` in git 

# TODO
- [x] Introduced `tags` column -> better grouping when visualizing results. Comma separated list of tags
- [x] Don't put expected tool calls and expected output in result csv - you can simply JOIN it from the input
- [x] Make `actual_output` hold the tool calls for tool related metrics
- [x] Introduce `{{ }}` meta logic to prevent the judge LLM from focusing on dynamic data
- [x] Use newest version of `deepeval` that has the fix of Tool Correctness
- [x] Add new cases
  - [x] literature search 1
- [x] Refactor
- [x] Make sure the default path is based on `__file__` and that the script can be run from anywhere basically
- [x] Somehow use tags?
- [x] If possible, create some clear standard output comparison of the `overall_results` update
- [x] Metrics and test cases don't correspond - seems like they is a mismatch
- [ ] Spit out the the exact command

### DIFF
--- a/backend/src/neuroagent/tools/autogenerated_types/entitycore.py
+++ b/backend/src/neuroagent/tools/autogenerated_types/entitycore.py
@@ -43,29 +43,6 @@ class AgePeriod(RootModel[Literal['prenatal', 'postnatal', 'unknown']]):
     root: Literal['prenatal', 'postnatal', 'unknown'] = Field(..., title='AgePeriod')
 
 
-class AnalysisNotebookExecutionAdminUpdate(BaseModel):
-    model_config = ConfigDict(
-        extra='allow',
-    )
-    authorized_public: bool | str | None = Field(
-        default='<NOT_SET>', title='Authorized Public'
-    )
-    start_time: AwareDatetime | str | None = Field(
-        default='<NOT_SET>', title='Start Time'
-    )
-    end_time: AwareDatetime | str | None = Field(default='<NOT_SET>', title='End Time')
-    used_ids: list[UUID] | str | None = Field(default='<NOT_SET>', title='Used Ids')
-    generated_ids: list[UUID] | str | None = Field(
-        default='<NOT_SET>', title='Generated Ids'
-    )
-    analysis_notebook_template_id: UUID | str | None = Field(
-        default='<NOT_SET>', title='Analysis Notebook Template Id'
-    )
-    analysis_notebook_environment_id: UUID | str | None = Field(
-        default='<NOT_SET>', title='Analysis Notebook Environment Id'
-    )
-
-
 class AnalysisNotebookExecutionCreate(BaseModel):
     model_config = ConfigDict(
         extra='allow',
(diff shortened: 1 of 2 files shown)


backend/eval/input/cerebellum_morphologies/expected_tool_calls.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "name": "entitycore-brainregion-getall",
+    "arguments": {
+      "name": "Cerebellum"
+    }
+  },
+  {
+    "name": "entitycore-reconstructionmorphology-getall",
```
WonderPG: The tool is now named `entitycore-cellmorphology-getall`

PR #487, opened 2025-08-21
Update obi-one morphometrics tool

### LINKED ISSUE #486: Let the LLM chose the requested features in the morphology feature tool
A recent update of obi-one allows us to specify the features we want to extract from the morphology feature tool.
One should autogen the new input and output schema from obi-one, add the input one as inheritance of the input of the current tool and double check that queries like 
```
Compute the soma size of the morphology
```
returns a much smaller output than before.

WonderPG:
Closes #486


jankrepl (approved):
Makes sense:) Thanks! @WonderPG

Issue #489, opened 2025-08-22
State sharing + POC on simulation config


jankrepl:
# Background
The idea here is to introduce some shared state between the backend (LLM) and the frontend that they can both `GET` and `PATCH`.

What exactly is a state? It is just a (potentially very nested) JSON

```json
{
    "simulation_config": {...}
}
```

For now the only key is `simulation_config` (POC for this PR) but in the future there will be other keys. The final goal of this PR is to allow for bidirectional state updates/fetches

* `backend-> frontend` When the `"obi-one-generatesimulationsconfig"` tool executes we want the frontend form to automatically update with the suggested changes (no copy pasting).
  * Note that we already have a similar mechanism in our frontend (also the official one) since vercel streams tool outputs so the frontend can literally  just check what tool was called and conditionally perform actions (e.g. display plot)
* `frontend->backend` When someone manually modifies the frontend form we want the LLM to have access to the latest state.
  * We have **a** mechanism that does that here: https://github.com/openbraininstitute/neuroagent/pull/433 . We force the frontend to send `frontend_url` with every request to `qa/chat_streamed`

# Inspiration

There seem to be some projects/frameworks that introduce the concept already. See an example below

* github repo: https://github.com/ag-ui-protocol/ag-ui
* tutorial video: https://youtu.be/0YnMZQ3UiSo?t=874

Now, in this PR I would maybe first think of doing this without a framework - but feel free to disagree. Also, the reason why I am saying this is that our backend is in Python and our frontend is in Next.js.

# TODO
- [ ] Read through docs of the `ag-ui` tool. Especially state management: https://docs.ag-ui.com/concepts/state. And if you think we should use the framework (or any other framework) instead of doing this from scratch speak up.
- [ ] (Optional) Adjust our frontend to demo this feature. Use `zustand` to store the whole state and maybe create a `/state` page that just shows the value. The goal is to show that calling the tool will lead to changes in the state ( `backend` -> `frontend` direction)
- [ ] Implement a POC from scratch. I could think of 2 general approaches and they differ in where the "ground truth" state is stored.

### Approach 1 (Frontend has the ground truth state)
TLDR it is basically what we do with the `frontend_url` but we also force the frontend to send us the whole `state` in the request body.

<img width="995" height="380" alt="Image" src="https://github.com/user-attachments/assets/daad0f5a-5b89-44ba-9352-9c8ebed1816e" />

### Approach 2 (Backend has the ground truth state)

[...]

PR #507, opened 2025-09-12
State sharing

WonderPG:
Contributes to #489 

# How it works

The backend is adjusted to accept descriptions of full simulation campaigns or incremental changes. The frontend passes the current simulation config in the request body which the backend recovers. The final simulation json is split into json patches that are returned in the tool output. The frontend is responsible for applying them to their state. 
The frontend shows a button next to relevant tools (smc simulation config in this case) which opens a sidebar. When the tool call is done, the JSON displayed updates. You can also modify it directly in the frontend, the changes will be seen by the backend on the next request

# Disclaimer
- I did not implement any logic to handle conflicts (frontend + backend changing the state simultaneously)
- I tried to implement a diffing mechanism but I gave up. The changes are all accepted by default. To see the diff mechanism in action, please checkout [a786bd0](https://github.com/openbraininstitute/neuroagent/pull/507/commits/a786bd02d8853b627f6093fb30c02578d2475a99) (no guarantee)
- The simulation state is thread bound (not stored in the store). I made this decision for simplicity. Each thread starts with a default simulation config, which can be updated through the tool. Changing thread sets your simulation config to the latest state it had in this conversation.
- I didn't add a global button to open the state. I estimated that the state is somewhat bound to the tool call, so you need to call the `obione-generatesimulationsconfig` to see the button appear and to be able to open the sidebar. It currently shows only the simulation config json, not the whole state.
- Because it is thread bound, when switching thread the simulation config defaults to a basic simulation, and all of the patches of the tools are applied sequentially. Therefore there are two issues
    - The changes made manually to the JSON in the frontend are not re-applied on reload
    - If you made a manual change then called the tool to make yet another modification that changes what you manually updated, it is possible that on reload the json won't be valid anymore (since your manual changes will not be applied, but the tool changes will)

The frontend is far from perfect, but tracking correctly this json and its diffs (tool + frontend) is too much of a task for this ticket imo. Feel free to disagree and make changes !

### DIFF
--- a/backend/src/neuroagent/tools/obione_generatesimulationsconfig.py
+++ b/backend/src/neuroagent/tools/obione_generatesimulationsconfig.py
@@ -81,6 +81,7 @@ Use this tool when users need to:
 
 Always call this tool to generate a simulation config, never attempt to generate one yourself.
 Do not try to generate a simulation config yourself if the tool fails.
+CRITICAL: DO NOT WRITE THE FINAL JSON OUTPUT IN THE CHAT. It will be automatically taken care of by the platform.
 """
     description_frontend: ClassVar[
         str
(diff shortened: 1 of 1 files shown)


jankrepl (approved):
LGTM:)

PR #508, opened 2025-09-15
Ts autogen

### LINKED ISSUE #264: Autogen: OpenAPI -> Typescript types and only use autogen types in the frontend
This will save us a lot of headache

https://openapi-ts.dev/introduction

I started this random convention that backend types would be prepended with `B` and then one would just try to rewrite Pydantic types to typescript types. However, this tool will literally autogen ALL possible types for request bodies, query params and and response bodies and just stores them in a single file. We can then directly import from that file.

WonderPG:
Closes #264


frontend/src/lib/neuroagent_types.ts
```diff
@@ -0,0 +1,3398 @@
+/**
```
jankrepl, line 1: IMO we could add a new job in the CI that checks that this file is up to date. The same logic as we have for the `/tools/autogenenerated_types/*`.

I am just afraid that unless we somehow enforce this file will not be up to date.


Also, any chance you can put the command to run in some `README.md`?

jankrepl (approved):

PR #515, opened 2025-09-17
Talk to circuit df

### LINKED ISSUE #500: "Talk to DataFrame" tool
# What 
A new tool that accepts the following two parameters
* `storage_id` - storage id of an asset that holds a serialized `pd.DataFrame` (probably csv)
* `query` - English description of how to query your dataframe

# How
### Option 1)
 It can just be a wrapper of this https://python.langchain.com/docs/integrations/tools/pandas/. Note that this tool uses the Python interpreter under the hood (probably runs some subprocess). 

> NOTE: this agent calls the Python agent under the hood, which executes LLM generated Python code - this can be bad if the LLM generated Python code is harmful. Use cautiously.

### Option 2)

https://python.langchain.com/docs/integrations/tools/sql_database/ In theory we would store an `.sqlite` in storage and then load it in memory and use the tool

### Option 3)
Probably there are more options. Feel free to research

# Comments
* Why storage id? It is currently the only way to store per user stuff without leaking it to the LLM
* How do I test it? Feel free to create anotehr tool that generates a random csv/.sqlite and saves it to storage. Or you can directly implement a tool that downloads a circuit from entitycore and then loads it via bluepysnap into pandas DataFrame.

BoBer78:
closes #500


BoBer78:
All in all the tool works, the only thing that might be left is to prompt it better. The issue is that for example, when asking for :
 
`What are the different e-types used in the circuit`
The LLM gets confused and sometimes tries to query the e-models with some strange parameters. 

Other than that the tool works fine on the benchmark provided by Jean-Denis (see the related issue)

backend/src/neuroagent/tools/circuit_population_analysis_tool.py
```diff
@@ -0,0 +1,335 @@
...
+                else:
+                    raise ValueError("No content in chat completion.")
+
+                # Clean up response (remove markdown if present)
+                if "```" in sql:
+                    parts = sql.split("```")
+                    for part in parts:
+                        if "SELECT" in part.upper():
+                            sql = part.replace("sql", "").strip()
+                            break
```
jankrepl, line 309: Why don't we use structured output instead? The current

```
```programming language
```
```

parsing is exactly how one would do things before structured output existed.

File: backend/src/neuroagent/tools/circuit_population_analysis_tool.py [hunk 2]
```diff
@@ -0,0 +1,335 @@
...
+            follow_redirects=False,
+        )
+        if response.status_code != 307:
+            raise ValueError(
+                f"The asset download endpoint returned a non 307 response code. Error: {response.text}"
+            )
+        presigned_url = response.headers["location"]
+
+        # Download the .gz file
+        print("DOWNLOADING FILE")
```
jankrepl, line 117: Let's use `logger.info` instead if really necessary

backend/src/neuroagent/tools/circuit_population_analysis_tool.py
```diff
@@ -0,0 +1,335 @@
...
+        os.makedirs(temp_dir, exist_ok=True)
```
jankrepl, line 118: A detail, throughout the codebase we use `pathlib` to do anything path related. I would stick to that convention.

backend/src/neuroagent/tools/circuit_population_analysis_tool.py
```diff
@@ -0,0 +1,335 @@
...
+
+        raise RuntimeError("Circuit population not found.")
+
+    def _is_safe_sql(self, sql: str) -> bool:
+        """Check if SQL is safe (only SELECT queries)."""
+        sql_upper = sql.upper().strip()
+        dangerous = ["DROP", "DELETE", "INSERT", "UPDATE", "CREATE", "ALTER", "EXEC"]
+        return sql_upper.startswith("SELECT") and not any(
+            word in sql_upper for word in dangerous
+        )
```
jankrepl, line 167: Can we do `@staticmethod` - it does not use `self`. Alternatively, put it outside the class.

backend/src/neuroagent/tools/circuit_population_analysis_tool.py
```diff
@@ -0,0 +1,335 @@
...
+            with tempfile.TemporaryDirectory() as temp_dir:
+                # Download and load the circuit population data
+                circuit_config_path = await self._download_and_extract_circuit(temp_dir)
+                population_dataframe = self._load_circuit_population_data(
+                    circuit_config_path
+                )
+                # population_dataframe = self._load_circuit_population_data("../nbS1-HEX1/circuit_config.json")
+
+                # Set up DuckDB connection
+                conn = duckdb.connect()
```
jankrepl, line 181: As discussed privately, I don't quite see why `duckdb` would be necessary here. AFAIK `sqlalchemy` would do the job. But happy to change my mind.

Also, no updates in `pyproject.toml`

jankrepl: I guess one convenient thing is that you can work directly with a pandas dataframe.

backend/src/neuroagent/tools/circuit_population_analysis_tool.py
WonderPG: Done in [1c0b27b](https://github.com/openbraininstitute/neuroagent/pull/515/commits/1c0b27bd30a5f7c48b22ae0f5b795f09c2e86eb5)

jankrepl (approved):

PR #518, opened 2025-09-19
Add PredefinedNodeSet in simulation config tool

### LINKED ISSUE #509: Add predefined nodesets to simulation config
The idea would be to call the following endpoint of obi-one. 

Image

For now there is no need to recursively look for the `circuit_property_type="CircuitNodeSet"` in the Pydantic types. However, at least make sure it is inside the autogen - currently it is not (hint check something like `--field-extra-keys` or `--field-extra-keys-without-x-prefix` - https://github.com/koxudaxi/datamodel-code-generator/discussions/414). Attention, it might not be in the `openapi.json` in the first place. 

Note that you need to use (probably locally) the following branch of obi-one https://github.com/openbraininstitute/obi-one/compare/predefined_nodeset_with_circuit_property

There are 2 different approaches:

1. We always do the `GET` and always give it to the LLM
3. We do a 2 step procedure. 1) Ask the LLM whether it wants the predefined nodesets for a given circuit id. And if it decides yes, then 2) `GET`. It is tricky.

WonderPG:
Closes #509


jankrepl (approved):
LGTM:)

PR #525, opened 2025-09-26
Remove sonata_asset_id dependency in circuit analysis tool

### LINKED ISSUE #524: Remove `asset_id` from the input of the circuit analysis tool
The `assert_id` can be inferred from a call to the `GET /circuit/{circuit_id}` endpoint of entitycore. We have observed the LLM messing up when providing this parameter so it is safer to maybe get the correct `asset_id` from the `circuit_id` ourselves inside of the tool.

WonderPG:
Closes #524


jankrepl (approved):
LGTM:)

PR #530, opened 2025-09-30
Specify base url via environment variable (necessary for Azure) + v0.9.3

### LINKED ISSUE #503: Support Azure OpenAI
The idea would be to somehow swap out our the current `AsyncOpenAI` for  `AsyncAzureOpenAI` in the dependency (based on some environment variable).

I am hoping that the two clients behave exactly the same way (e.g. structured output, chat completions, ...). But if not we will also have to somehow adjust our code

jankrepl:
Closes #503


WonderPG (approved):
As discussed privately I would maybe keep open the possibility to pass the openrouter client in the tools through the context variables, but we can do it in another PR.

PR #532, opened 2025-10-01
Fix question suggestions

### LINKED ISSUE #529: question_suggestion behaves opposite to what is should
It seems like even though the frontend request is correct, the LLM is considering the click history in reverse order. For instance:

Image

or

Image

or

Image

Image

The last element should be the "most important one" but currently it is the first it seems.

BoBer78:
closes #529


backend/src/neuroagent/app/routers/qa.py
```diff
@@ -161,58 +162,60 @@ async def question_suggestions(
...
+- Artifacts may include:
+  * Morphology
+  * Electrophysiology
+  * Neuron density
+  * Bouton density
+  * Synapse per connection
+  * E-model
+  * ME-model
+  * Synaptome
+- The current date and time is {datetime.now(timezone.utc).isoformat()}. Weight the user clicks depending on how old they are. The more recent clicks should be given a higher importance.
```
WonderPG, line 197: I suggest we put the date towards the end to maximize prompt caching

backend/src/neuroagent/app/routers/qa.py
```diff
@@ -161,58 +162,60 @@ async def question_suggestions(
...
+
+Task:
+{"Using either the user's navigation history or their recent messages, generate three short, literature-focused questions they might ask next." if is_in_chat else "Based on the user's navigation history, generate a short, literature-focused question about the last brain region they visited."}
```
WonderPG, line 200: Above we have the following if/else statement
```
if is_in_chat:
        content = f"CONVERSATION MESSAGES: \n{json.dumps([{k: v for k, v in json.loads(msg.content).items() if k in ['role', 'content']} for msg in db_messages])}"
    else:
        content = (
            f"USER JOURNEY: \n{body.model_dump(exclude={'thread_id'})['click_history']}"
        )
```
i.e. we either use the user journey or the conversation history, no inbetween. So if `in_in_chat`, we shouldn't mention user journey.

WonderPG (commented):
Thanks for the PR ! Two small comments
- Line 159 could we add `mode="json"` as a `model_dump` argument to make sure the time format is the same in the prompt and in the user message ?
- I would maybe double check with Fabien if it is okay to return a list with one element for the first suggestion and 3 afterward.

WonderPG (approved):

PR #533, opened 2025-10-01
Switch to flash for tool selection

WonderPG:


jankrepl (commented):
I confirm that it is way faster than `main`.


To be discussed privately whether we actually merge this or not.

jankrepl (approved):

PR #537, opened 2025-10-01
Adapt to connections not closing anymore

### LINKED ISSUE #534: Adapt to fastapi 0.118.0
In this new version, they finally fixed the bug that closed the `yield` dependencies when hitting the `StreamingResponse` https://github.com/fastapi/fastapi/releases/tag/0.118.0. This means that in theory we don't need `stream.py` anymore to reinstantiate the dependencies. Our code needs a little refactor to be compatible.

WonderPG:
Closes #534


BoBer78 (approved):
Works great, farewell `stream.py` you won't be missed 🦫

PR #545, opened 2025-10-07
Try to enforce using metric tools rather than downloading assets

### LINKED ISSUE #543: LLM suggests asset downloading instead of using metrics tools
It happens sometimes that the LLM starts suggesting that you download an asset belonging to some entity (e.g. nwb asset of an electrical cell recording) and then somehow suggest one needs to download it in order to compute metrics.

**this is wrong**

The morpho/electrophys/circuit metrics only need the identifier of the entitycore entity. Nothing else.

Example conversation of when things got wrong

Image

Reported by @ayimaok

WonderPG:
Closes #543 I guess ? No idea if that solves it for good.


jankrepl (approved):
LGTM:)

jankrepl (approved):

PR #554, opened 2025-10-15
bring back images in Literature Search

### LINKED ISSUE #548: Make sure we insert exa images in literature

### LINKED ISSUE #558: Make `morphology_studies` test case pass
https://github.com/openbraininstitute/neuroagent/blob/ff3504c3ad13269874791092e8f26ecd632fe892/backend/eval/output/scores.json#L17-L21

Or the more detailed overview

https://github.com/openbraininstitute/neuroagent/blob/ff3504c3ad13269874791092e8f26ecd632fe892/backend/eval/output/detailed.json#L179-L241

TLDR it tries to find the information in entitycore. Of course not sure how reproducible this is.

BoBer78:
closes #548, closes #558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Some prompt engineering.
 - Output class of circuit population tool.
+- Tool argument correctness metric.
 
 ## Changed
 - Running deepeval github action also on main
--- a/backend/eval/input/cerebellum_morphologies/expected_tool_calls.json
+++ b/backend/eval/input/cerebellum_morphologies/expected_tool_calls.json
@@ -2,7 +2,7 @@
   {
     "name": "entitycore-brainregion-getall",
     "arguments": {
-      "name": "Cerebellum"
+      "semantic_search": "Cerebellum"
     }
   },
   {
--- a/backend/eval/input/platform_news/user.md
+++ b/backend/eval/input/platform_news/user.md
@@ -1,1 +1,1 @@
-Please get me some news about the platform
+Please get me some news about the platform.
--- a/backend/eval/input/thalamus_id/expected_tool_calls.json
+++ b/backend/eval/input/thalamus_id/expected_tool_calls.json
@@ -2,7 +2,7 @@
   {
     "name": "entitycore-brainregion-getall",
     "arguments": {
-      "name": "Thalamus"
+      "semantic_search": "Thalamus"
     }
   }
 ]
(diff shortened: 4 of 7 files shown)


backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -1,29 +1,37 @@
...
-- **NEVER** fall back to name searches when you already have the ID in memory.
-- **INVALID PARAMETER**: `brain_region__id` does not exist and must not be used.
+## Brain Region ID Resolution
 
-## GET ALL Endpoint Guidelines
-- When calling `GET ALL` endpoints, please mention the `total_items` count and ask the user if they want to see additional data before using the `pagination` parameter to retrieve subsequent pages.
-- If a `GET ALL` endpoint returns an empty asset list, they can be accessed through the corresponding `GET ONE` endpoint or `GET ALL Assets` if needed.
+**MANDATORY workflow:**
+1. Resolve names to IDs using `entitycore-brainregion-getall` with `name` parameter (capitalize first letter)
+2. **ALWAYS use `within_brain_region_brain_region_id` with UUID** for subsequent queries
```
WonderPG: I'm wondering if this one is still useful. We might as well disable any other option in the existing params.

backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -1,29 +1,37 @@
...
+- EntityCore database only stores scientific names
 
-## Species request
-When a user requests data for a species using a common name:
-- Always use the scientific name when tools allow for specifying a species name. The Entitycore database only holds the scientific name of species.
-- Use the tool `entitycore-species-getall` to lookup species ID if needed.
+## GET ALL Endpoints
+- Mention `total_items` count in response
+- Ask before using pagination to retrieve additional pages
+- If empty asset list, use `GET ONE` endpoint or `GET ALL Assets`
```
WonderPG: iirc this one is not useful anymore. I believe all of the `GET ALL Entity` tools return the available assets now.

backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -1,29 +1,37 @@
...
+
+## Asset URLs
+- **ONLY `entitycore-asset-downloadone` returns valid URLs**
+- Other endpoints return invalid URLs - do not display them
+- Always include entity links using `url_link` field
```
WonderPG: Here we have a bit of a contradiction. We say that only `entitycore-asset-downloadone` returns valid urls, and that `url_link` should be included for every entity. It is not true anymore that only `entitycore-asset-downloadone` returns valid urls since we inject the frontend url inside of some entities.

backend/src/neuroagent/rules/entitycore-tools.mdc
jankrepl: AFAIK we always include all the `assets` in all the `GET ONE` and `GET ALL`  so forcing the LLM to do an extra call does not make sense

PR #555, opened 2025-10-16
Plot through Python tool

WonderPG:
To use packages that are not built in pyodide packages, please download the wheel manually and move it to a folder called`./backend/cached_wheels` (e.g. plotly)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,16 +51,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tool argument correctness metric.
 - Fix brain region get all for new hierarchy.
 - Images not appearing in Literature search (Prompt refactor).
+- Update uv lock package.
 
 ### Changed
 - Running deepeval github action also on main
 - Use typescript autogen for backend types in frontend.
 - Try to enforce using metric tools rather than downloading assets.
 - Rule to avoid overvalidating.
-
-## Fixed
-- Images not appearing in Literature search (Prompt refactor).
-- Update uv lock package.
 
 ## [v0.10.0] - 2.10.2025
 


backend/src/neuroagent/executor.py
```diff
@@ -0,0 +1,312 @@
...
+            with open(runner_path, "w") as f:
+                f.write(
+                    self.JS_CODE.format(
+                        packages=self.additional_imports, code=json.dumps(code)
+                    )
+                )
+
+            cmd = [self.deno_path, "run"] + self.deno_permissions + [runner_path]
+
+            # Start the server process
```
jankrepl: AFAIK there is no `deno serve` anymore. If that is the case then this comment is misleading.

backend/src/neuroagent/executor.py
```diff
@@ -0,0 +1,312 @@
...
+            events = []
+            if process.stdout:
+                for line in process.stdout:
+                    events.append(line.strip())
+                    if self.logger:
+                        # Funnel the logging to your logger as a debug log.
+                        self.logger.debug(line)
+
+            # Wait for the process to finish or timeout
+            process.wait(timeout=self.timeout)
```
jankrepl: So this is blocking. I know that this code was mostly taken from `smolagents`. I haven't benchmarked things to see how slow/fast it is and what happens when multiple requests happen at the same time.

As opposed to `smolagents`, `run-mcp-python` no server gets created. Instead, every tool call = new suprocess so IMO this `process.wait` could become a blocker.


Async alternative https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_exec

WonderPG: This actually comes from `mcp_run_python` which claims being async https://github.com/pydantic/mcp-run-python/blob/48805ec406261dcdd086e78bcd048b274120085a/mcp_run_python/main.py#L114.  If it's low hanging I will modify it though.

As you mentioned the `__enter__` user `subprocess.run` which is definitely blocking because it is ran only at startup.

backend/eval/input/plotting/user.md
```diff
@@ -1 +1 @@
-Can you please generate a plot y=x^2 with matplotlib?
+Can you please generate a plot of y=x^2?
```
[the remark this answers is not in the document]
WonderPG: I renamed the folder and updated the output. I did not re-run the eval though

jankrepl (dismissed):
LGTM:) I experimented locally and it works great. The WasmExecutor is complex but I guess I get most of the ideas. 

Great job @WonderPG

BoBer78 (approved):
Appart from a small question about the CSS file, Looks good to me ! 
I tested it locall and the plots are Amazing. Thanks @WonderPG !! 🦫

PR #557, opened 2025-10-20
Vercel v5

### LINKED ISSUE #352: Handle infinite render error
Currently, developing using `npm run dev` is very painful. We get these "Infinite render" issues all the time. It has gotten worse over time. Not blaming anyone - we should have been careful about this when adding all the  features to `main`.

Image

AFAIK this comes from the `useChat` being triggered way to often. One can reproduce this by plotting for example.

### LINKED ISSUE #566: Adapt to response API
Openrouter is now compatible with the response API https://openrouter.ai/docs/api-reference/responses-api/overview

We therefore have no reason left to stick to chat_completion. Among the benefits of the response API, we would get the reasoning summary of gpt-5 to display in our frontend (please make sure it works) as well as potential built in tools (web-search, python execution etc...)

https://platform.openai.com/docs/guides/migrate-to-responses

### LINKED ISSUE #580: Refactor HIL - tool input modification not allowed
Currently, when a HIL tool is called the user can take 3 different actions

1. Reject (+ optionally give a reason)
2. Accept without modifying the tool call input
3. Accept and modify the tool call input

Inspired by Cursor and Claude Desktop, let's get rid of `3. Accept and modify the tool call input`.
 IMO this will significantly reduce the complexity. Also, take the time to refactor the logic to make it as simple as possible to achieve the 1. + 3.

BoBer78:
Closes #580 closes #566 and closes #352 

# Main things 

- Made the switch to the Response API.
- The `Messages` DB schema is now completely different --> mimics the ResponseAPI. 
- Switch to Vercel v5 --> has same schema as Response API.
- Refactored HIL to only allow for Accept / Reject.
- Fixed some frontend bugs.

## DB Schema change 

IMPORTANT : if you want to backup your local DB before running `alembic upgrade`, I used : 

```bash
docker exec -i  <CONTAINER_ID>  pg_dump -U postgres -d neuroagent -F c -v > backup.dump

docker exec -i <CONTAINER_ID>   pg_restore -U postgres -d neuroagent -v --clean < backup.dump
```
I only kept the mandatory fields so that the we can send the messages back to OpenAI. Some non mandatory info is lost in the process.

The DB messages go from `User`, `AI_TOOL`, `TOOL` and `AI_MESSAGE` to only `User` and `Assistant`. Everything is then stored in the `Parts` of the messages : 
- `User` messages only have one `Message` part
- `Assistant` can have `Reasoning`, `Tool_call`, `Tool_call_output` and `Message` parts (Any number in any order)
One `Assistant` message now correspond to one full turn of the agent loop in AgentRoutine. 

The most complex part of the PR lies the [migration script](https://github.com/openbraininstitute/neuroagent/blob/f6250e7cb16586d049e05bb9e5b72ea08b23c1a0/backend/alembic/versions/25cefa8449c6_change_to_response_api.py). 

I tested extensively. You can write some messages, upgrade to new schema, write some more messages, downgrade, and repeat 10 times. It works. I did not test all the weird edge case like for example stopping a HIL message at a very specific timing. 

## Reponse API change

IMPORTANT : The response API of OpenRouter is still in Beta. When not streaming, it works fine, but they send the chunks in a different order sometimes and the main LLM is not compatible with it for now. 
For non streamed responses or parsed structured output it works fine. 

With the new change DB schema, the only thing we have to do is : 
https://github.com/openbraininstitute/neuroagent/blob/f6250e7cb16586d049e05bb9e5b72ea08b23c1a0/backend/src/neuroagent/utils.py#L28C1-L40C27 
And we can send it to OpenAI. 

For reference, here is what we send to the completion API : 
```
[
## First USER MESSAGE
{'content': 'Give me the ID of the thalamus ', 'role': 'user'},
 
## AI MAKING TOOL CALL
 {'content': '',
  'function_call': None,
  'reasoning': '',
  'role': 'assistant',
  'sender': 'Agent',
  'tool_calls': [{'function': {'arguments': '{"semantic_search":"thalamus","page":1,"page_size":10}',
                               'name': 'entitycore-brainregion-getall'},
                  'id': '2ab1d71ee7c344f6a304b94453ddcb00',
                  'type': 'function'}]},

# EXECUTING THE TOOL
 {'content': 'TEST OUTPUT',
  'role': 'tool',
  'tool_call_id': '2ab1d71ee7c344f6a304b94453ddcb00',
  'tool_name': 'entitycore-brainregion-getall'},

# FINAL LLM CONTENT
 {'content': 'TEST CONTENT',
  'function_call': None,
  'reasoning': '',
  'role': 'assistant',
  'sender': 'Agent',
  'tool_calls': None},

# NEXT USER MESSAGE
 {'content': 'dour', 'role': 'user'}
  ]
```
And here is what we send to the Reponse API : 
```
[
    
## First USER MESSAGE
{'content': [{'text': 'Give me the ID of the thalamus ',
               'type': 'input_text'}],
  'role': 'user',
  'status': 'completed',
  'type': 'message'},

## AI MAKING TOOL CALL
 {'content': [],
  'encrypted_content': 'SECRET REASONING',
  'summary': [],
  'type': 'reasoning'},

 {'arguments': '{"semantic_search": "thalamus", "page": 1, "page_size": 10, '
               '"id": null, "acronym": null, "annotation_value": null, '
               '"hierarchy_id": "e3e70682-c209-4cac-a29f-6fbed82c07cd", '
               '"order_by": ["name"]}',
  'call_id': 'd362cd7db6c1414e99b3ab1cb6b4a527',
  'name': 'entitycore-brainregion-getall',
  'status': 'completed',
  'type': 'function_call'},

# EXECUTING THE TOOL
 {'call_id': 'd362cd7db6c1414e99b3ab1cb6b4a527',
  'output': 'TEST OUTPUT',
  'status': 'completed',
  'type': 'function_call_output'},

# FINAL LLM CONTENT
 {'content': [],
  'encrypted_content': 'SECRET REASONING',
  'summary': [{'text': 'Summary 1',
               'type': 'summary_text'}],
  'type': 'reasoning'},

 {'content': [{'text': 'TEST CONTENT'}],
  'role': 'assistant',
  'status': 'completed',
  'type': 'message'},

# NEXT USER MESSAGE
 {'content': [{'text': 'dour', 'type': 'input_text'}],
  'role': 'user',
  'status': 'completed',
  'type': 'message'}
  ]
  ```

## AgentRoutine Change 

I made everything work with the pydantic schemas from OpenAI. So now there is no random dict, everything is an OpenAI type. One very conveniant change is that OpenAI now does the concatenation on their side. Once a `Part` is finished streaming, OpenAI sends us a chunk with the complete part. 

https://github.com/openbraininstitute/neuroagent/blob/f6250e7cb16586d049e05bb9e5b72ea08b23c1a0/backend/src/neuroagent/agent_routine.py#L293 : this keeps the temporary data if the user stop the stream. It also keeps the current tool call that need to be executed until they are done. 

The new parts are appended (when they are complete) both in the history and the new_message with this function : 
https://github.com/openbraininstitute/neuroagent/blob/f6250e7cb16586d049e05bb9e5b72ea08b23c1a0/backend/src/neuroagent/utils.py#L250C1-L272C27

## Vercel v5

A lot of small things were changed. The main annoyance was the the way the chunks were streamed changed completely. 
But with the switch to the response API, one vercel chunk == one Response chunk so it was not that complicated.

### DIFF
--- a/backend/src/neuroagent/app/dependencies.py
+++ b/backend/src/neuroagent/app/dependencies.py
@@ -17,6 +17,7 @@ from openai import AsyncOpenAI
 from redis import asyncio as aioredis
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
+from sqlalchemy.orm import selectinload
 from starlette.status import HTTP_401_UNAUTHORIZED
 
 from neuroagent.agent_routine import AgentsRoutine
@@ -518,6 +519,7 @@ async def filtered_tools(
     filtered_models: Annotated[
         list[OpenRouterModelResponse], Depends(get_openrouter_models)
     ],
+    session: Annotated[AsyncSession, Depends(get_session)],
 ) -> tuple[list[type[BaseTool]], dict[str, str | None]]:
     """Based on the current conversation, select relevant tools."""
     if request.method == "GET":
@@ -528,10 +530,15 @@ async def filtered_tools(
 
     body = await request.json()
 
-    messages: list[Messages] = await thread.awaitable_attrs.messages
+    thread_result = await session.execute(
+        select(Threads)
+        .options(selectinload(Threads.messages).selectinload(Messages.parts))
+        .where(Threads.thread_id == thread.thread_id)
+    )
 
-    for message in messages:
-        await message.awaitable_attrs.parts
+    # Re-assign the thread so that they are loaded for astream.
+    thread = thread_result.scalar_one()
+    messages = thread.messages
 
     if (
         not messages

### CODE REVIEW

File: backend/src/neuroagent/agent_routine.py [hunk 1]
```diff
@@ -52,35 +64,39 @@ async def get_chat_completion(
...
             else agent.instructions
         )
-        messages = [{"role": "system", "content": instructions}] + history
 
         tools = [tool.pydantic_to_openai_schema() for tool in agent.tools]
 
         create_params = {
-            "messages": messages,
+            "instructions": instructions,
+            "input": history,
```
WonderPG: If I am seeing it correctly, https://platform.openai.com/docs/api-reference/responses/create states that `input` should be used for the very last message only (i.e. a string) while the remaining conversation should be passed under `conversation`. I might be wrong though please double check.

frontend/src/components/chat/chat-page.tsx
```diff
@@ -75,35 +79,47 @@ export function ChatPage({
...
-      return {
-        content: lastMessage.content,
-        tool_selection: selectedTools,
-        model: currentModel.id,
-        frontend_url: frontendUrl,
-      };
-    },
+    messages: retrievedMessages,
+    experimental_throttle: 50,
+    sendAutomaticallyWhen: lastAssistantHasAllToolOutputs,
```
BoBer78: Now we have to specify when we want the agent to send a message automatically.

frontend/src/components/chat/chat-page.tsx
```diff
@@ -151,43 +163,26 @@ export function ChatPage({
...
-            .at(-1)
-            ?.annotations?.map((ann) =>
-              !ann.toolCallId ? { isComplete: false } : ann,
-            ),
-        };
-        // We only change the annotation at message level and keep the rest.
-        return prevState;
-      });
-    }
-  }, [stopped, setMessages]);
```
BoBer78: Might want to revisit that ... it works for now.

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                                    status="in_progress",
+                                )
+                            )
+                            yield f"data: {json.dumps({'type': 'start-step'})}\n\n"
+                            yield f"data: {json.dumps({'type': 'tool-input-start', 'toolCallId': event.item.call_id, 'toolName': event.item.name})}\n\n"
+
+                        # Tool call (args) deltas
+                        case ResponseFunctionCallArgumentsDeltaEvent() if event.item_id:
+                            # Now call ID in the stream, we have to get it form temp stream data
+                            tool_call_id = temp_stream_data["tool_calls"][
```
BoBer78: For some reason, OpenAI does not stream the `call_id` in the deltas. 

Since the `call_id` is the only thing we have in the old DB schema before migration, and is the ONLY mandatory part in the OpenAi ReponseAPI to identify tool calls, I made the decision to keep the `call_id` as the main way to identify the tool call. 

The frontend works only with this `call_id`, nothing else. Feel free to challenge that decision.

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                            yield f"data: {json.dumps({'type': 'finish-step'})}\n\n"
+
+                        # === Usage ===
+                        # Handle usage/token information and ecrypted reasoning.
+                        case ResponseCompletedEvent():
+                            usage_data = event.response.usage
+
+                        # case _:
+                        #     print(event.type)
+                        # Some events are not needed. Not sure what we should do with them yet.
```
BoBer78: Some cases were not taken into account since they where duplicates. For example the "True Reasoning" and not the summaries may appear in other event types. Since OpenAI does not show them I am not sure how to test it ...

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                yield f"data: {json.dumps({'type': 'finish', 'messageMetadata': {'toolCalls': metadata_data}})}\n\n"
+            else:
+                yield f"data: {json.dumps({'type': 'finish'})}\n\n"
+            yield "data: [DONE]\n\n"
 
         # User interrupts streaming
         except asyncio.exceptions.CancelledError:
-            if isinstance(message["tool_calls"], defaultdict):
-                message["tool_calls"] = list(message.get("tool_calls", {}).values())
+            # add parts not appended to `new_message`
```
BoBer78: We try to append the missing parts to new_messages with `is_complete = False`. Then we simply append the new_message and the background task will append it to the DB.

backend/alembic/versions/25cefa8449c6_change_to_response_api.py
```diff
@@ -0,0 +1,935 @@
...
+            except:
+                content_json = {"content": content}
+
+            if entity == "USER":
+                # Create USER message with one MESSAGE part
+                user_text = content_json.get("content", "")
+                conn.execute(
+                    sa.text("""
+                    INSERT INTO parts (part_id, message_id, order_index, type, output, is_complete, validated)
+                    VALUES (gen_random_uuid(), :message_id, 0, 'MESSAGE', :output, :is_complete, NULL)
```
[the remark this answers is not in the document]
WonderPG: Same later

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                        # Tool call starts
+                        case ResponseOutputItemAddedEvent() if (
+                            isinstance(event.item, ResponseFunctionToolCall)
+                            and event.item.id
+                        ):
+                            temp_stream_data["tool_calls"][event.item.id] = (
+                                ResponseFunctionToolCall(
+                                    id=event.item.id,
+                                    call_id=event.item.call_id,
+                                    name=event.item.name,
```
WonderPG: We used to generate random UUID ourselves to be compatible with every provider (openrouter). Is it a problem to redefine a UUID here ?

backend/src/neuroagent/agent_routine.py
```diff
@@ -52,35 +67,39 @@ async def get_chat_completion(
         context_variables: dict[str, Any],
         model_override: str | None,
         stream: bool = False,
-    ) -> AsyncStream[ChatCompletionChunk]:
+    ) -> AsyncStream[ResponseStreamEvent]:
```
WonderPG: This function should be renamed.

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                            ] += event.delta
+                            yield f"data: {json.dumps({'type': 'reasoning-delta', 'id': event.item_id, 'delta': event.delta})}\n\n"
+
+                        # Reasoning summary end
+                        case ResponseReasoningSummaryPartDoneEvent():
+                            yield f"data: {json.dumps({'type': 'reasoning-end', 'id': event.item_id})}\n\n"
+                            yield f"data: {json.dumps({'type': 'finish-step'})}\n\n"
+
+                        # Capture the final reasoning (only chunk that has the encrypted content)
+                        case ResponseOutputItemDoneEvent() if (
```
WonderPG: The `ResponseOutputItemDoneEvent` could also be unified imo since all of the part type share the same logic.

backend/src/neuroagent/app/app_utils.py
```diff
@@ -578,15 +503,20 @@ async def filter_tools_and_model_by_conversation(
         # Send the OpenAI request
         model = "google/gemini-2.5-flash"
         start_request = time.time()
-        response = await openai_client.beta.chat.completions.parse(
-            messages=[{"role": "system", "content": system_prompt}, *openai_messages],  # type: ignore
+        response = await openai_client.responses.parse(
+            input=[{"role": "system", "content": system_prompt}, *filtered_messages],  # type: ignore
```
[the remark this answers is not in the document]
WonderPG: This is very weird. Worth an investigation

backend/src/neuroagent/app/database/sql_schemas.py
```diff
@@ -25,12 +26,19 @@ def utc_now() -> datetime.datetime:
...
+    ASSISTANT = "assistant"
+
+
+class PartType(enum.Enum):
+    """Type of Response API part."""
+
+    MESSAGE = "message"
+    REASONING = "reasoning"
+    FUNCTION_CALL = "function_call"
+    FUNCTION_CALL_OUTPUT = "function_call_output"
```
WonderPG: As raised above, I doubt the need of this column/enum

backend/src/neuroagent/app/database/sql_schemas.py
```diff
@@ -123,27 +133,27 @@ class Messages(Base):
...
-    message: Mapped[Messages] = relationship("Messages", back_populates="tool_calls")
+    order_index: Mapped[int] = mapped_column(Integer, nullable=False)
+    type: Mapped[PartType] = mapped_column(Enum(PartType), nullable=False)
+    output: Mapped[dict[str, Any]] = mapped_column(JSONB, nullable=False)
+    is_complete: Mapped[bool] = mapped_column(Boolean, nullable=False)
+    validated: Mapped[bool] = mapped_column(Boolean, nullable=True)
+
+    message: Mapped[Messages] = relationship("Messages", back_populates="parts")
+
+    __table_args__ = (Index("ix_parts_message_id", "message_id"),)
```
WonderPG: Shouldn't this be on the `Messages` table ?

backend/src/neuroagent/app/dependencies.py
```diff
@@ -520,19 +528,36 @@ async def filtered_tools(
 
     body = await request.json()
 
-    # Awaiting here makes downstream calls already loaded so no performance issue
     messages: list[Messages] = await thread.awaitable_attrs.messages
+
+    for message in messages:
+        await message.awaitable_attrs.parts
```
WonderPG: To avoid X round trips to the db (thread + messages + (X-2) parts) I suggest we don't use the `get_thread` dependency here and simply get it from the session while lazy loading messages and parts.
Something along the lines of:
```
if not thread_id:
        raise HTTPException(
            status_code=400,
            detail={"error": "thread_id is missing"},
        )
    
    thread = await session.get(
        Threads, 
        thread_id, 
        options=[selectinload(Threads.messages).selectinload(Messages.parts)]
    )
    if not thread:
        raise HTTPException(
            status_code=404,
            detail={"error": "Thread not found"},
        )
```

backend/src/neuroagent/app/routers/tools.py
```diff
@@ -37,70 +40,89 @@
 
 @router.patch("/{thread_id}/execute/{tool_call_id}")
 async def execute_tool_call(
-    thread_id: str,
+    thread_id: UUID,
```
WonderPG: This is unused.

backend/src/neuroagent/app/database/sql_schemas.py
```diff
@@ -123,27 +133,27 @@ class Messages(Base):
...
+    __tablename__ = "parts"
+    part_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, primary_key=True, default=lambda: uuid.uuid4()
+    )
     message_id: Mapped[uuid.UUID] = mapped_column(
-        UUID, ForeignKey("messages.message_id")
+        UUID, ForeignKey("messages.message_id"), nullable=False
     )
-    message: Mapped[Messages] = relationship("Messages", back_populates="tool_calls")
+    order_index: Mapped[int] = mapped_column(Integer, nullable=False)
```
WonderPG: iirc it is possible to make a field auto-increment. Just an idea.

backend/src/neuroagent/tools/base_tool.py
```diff
@@ -166,29 +166,26 @@ class BaseTool(BaseModel, ABC):
...
+        return {
             "type": "function",
             "name": cls.name,
             "description": cls.description,
-            "function": {
-                "name": cls.name,
-                "description": cls.description,
-                "strict": False,
-                "parameters": parameters,
-            },
```
WonderPG: Is this solved in the response API ? I would imagine it isn't

frontend/src/components/chat/chat-messages-inside-thread.tsx
```diff
@@ -109,7 +119,7 @@ export function ChatMessagesInsideThread({
             })}
           </div>
         ) : (
-          <ChatMessageHuman key={message.id} content={message.content} />
+          <ChatMessageHuman key={message.id} content={getLastText(message)} />
```
WonderPG: What is the purpose of `getLastText` ? We are in the case where `message.role === "user"` aren't we guaranteed that `message.part.text` exists ?

frontend/src/components/chat/chat-page.tsx
```diff
@@ -75,35 +79,48 @@ export function ChatPage({
...
+      api: `${env.NEXT_PUBLIC_BACKEND_URL}/qa/chat_streamed/${threadId}`,
+      headers: {
+        Authorization: `Bearer ${session?.accessToken}`,
+      },
+      prepareSendMessagesRequest: ({ messages }) => {
+        const checkedToolsNow = useStore.getState().checkedTools;
+        const currentModelNow = useStore.getState().currentModel;
+        return {
+          body: {
+            content: getLastMessageText(messages),
```
WonderPG: Why is this different than before ? To my understanding, content is simply `lastMessage.parts.at(-1).text` no ?

backend/src/neuroagent/app/app_utils.py
```diff
@@ -578,15 +503,20 @@ async def filter_tools_and_model_by_conversation(
...
             model=model,
-            response_format=ToolModelFiltering,
+            text_format=ToolModelFiltering,
+            store=False,
         )
 
         # Parse the output
-        if response.choices[0].message.parsed:
-            parsed = response.choices[0].message.parsed
+        parsed = response.output_parsed
```
WonderPG: `if need_model_selection` is wrongly nested. It should be independent of the `need_tool_selection`. Currently it is not

backend/alembic/versions/25cefa8449c6_change_to_response_api.py
```diff
@@ -0,0 +1,935 @@
...
+                """),
+                    {
+                        "message_id": msg_id,
+                        "output": json.dumps(
```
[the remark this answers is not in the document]
BoBer78: So I tried removing it, and I got this error : `sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) can't adapt type 'dict'`

From Claude : "You need to use json.dumps() when using raw SQL with sa.text() because psycopg2 can't adapt dict types directly"

backend/alembic/versions/25cefa8449c6_change_to_response_api.py
```diff
@@ -0,0 +1,935 @@
...
+    op.drop_index("ix_messages_search_vector", table_name="messages", if_exists=True)
+
+    # Create Parts table
+    op.create_table(
+        "parts",
+        sa.Column("part_id", sa.UUID(), nullable=False),
+        sa.Column("message_id", sa.UUID(), nullable=False),
+        sa.Column("order_index", sa.Integer(), nullable=False),
+        sa.Column(
+            "type",
```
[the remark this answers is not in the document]
BoBer78: Is was more for convenience. It is very practical to when you look at the DB in DBober. 

I can remove it, but it will require a small rework of the PR

backend/src/neuroagent/agent_routine.py
```diff
@@ -132,23 +151,24 @@ async def execute_tool_calls(
...
-            return {
-                "role": "tool",
-                "tool_call_id": tool_call.tool_call_id,
-                "tool_name": name,
-                "content": f"Error: Tool {name} not found.",
-            }, None
+            return ResponseFunctionToolCallOutputItem(
+                id=tool_call.id or "",
+                call_id=tool_call.call_id,
+                status="incomplete",
```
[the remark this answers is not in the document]
BoBer78: I did not check this one. We can investigate more.

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                            yield f"data: {json.dumps({'type': 'reasoning-delta', 'id': event.item_id, 'delta': event.delta})}\n\n"
+
+                        # Reasoning summary end
+                        case ResponseReasoningSummaryPartDoneEvent():
+                            yield f"data: {json.dumps({'type': 'reasoning-end', 'id': event.item_id})}\n\n"
+                            yield f"data: {json.dumps({'type': 'finish-step'})}\n\n"
+
+                        # Capture the final reasoning (only chunk that has the encrypted content)
+                        case ResponseOutputItemDoneEvent() if (
+                            isinstance(event.item, ResponseReasoningItem)
```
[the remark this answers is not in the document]
BoBer78: So there are a lot of different encrypted reasoning chunks.
There is : 
- The creation of the Summary part
- The Finish step of the part
- The last chunk of the Complete response. This yet again has a different encrypted content. 

From what I read, the encrypted content to store and send back is the one that is associated to the complete part, the way I am doing right now.

backend/src/neuroagent/app/app_utils.py
```diff
@@ -253,147 +238,84 @@ def format_messages_vercel(
...
-                id=uuid.uuid4(),
-                role="assistant",
-                createdAt=msg.creation_date,
-                content="",
-                parts=parts,
-                annotations=annotations,
-            )
+            elif part.type == PartType.FUNCTION_CALL_OUTPUT:
+                tc_id = output.get("call_id", "")
+                if tc_id in tool_calls:
```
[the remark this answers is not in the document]
BoBer78: Indeed, I will raise an error when it is the case.

backend/src/neuroagent/app/app_utils.py
```diff
@@ -505,10 +427,13 @@ async def filter_tools_and_model_by_conversation(
...
     openai_messages = await messages_to_openai_content(messages)
 
-    # Remove the content of tool responses to save tokens
-    for message in openai_messages:
-        if message["role"] == "tool":
-            message["content"] = "..."
+    filtered_messages = []
+    for msg in openai_messages:
+        if msg.get("type") == PartType.REASONING.value:
+            continue
```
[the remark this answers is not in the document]
BoBer78: I think they where not compatible when I tested the Response API of OpenRouter. I would not include them since we are using different models / providers for the Main LLM and the tool selector.

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
-                                    yield f"c:{json.dumps(args_data, separators=(',', ':'))}\n"
-                                    draft_tool_calls[draft_tool_calls_index][
-                                        "arguments"
-                                    ] += arguments
-                        elif (
-                            hasattr(choice.delta, "reasoning")
-                            and choice.delta.reasoning
+                usage_data = None
+                # for streaming interrupt and handling tool calls
+                temp_stream_data: dict[str, Any] = {
```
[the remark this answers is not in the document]
BoBer78: Why I went with this approach is that to me it made more sense to only take the `finished` parts from OpenAI and put it straight in the DB. I added the `temp_stream_data` afterwards to handle the streaming. 

Since I was not sure if concurrent parts could be opened or not, I kept everything separate. 
Edit: I tested, even in the case of parrallel tool calls, the parts are always closed first before a new one is opened. So yes it will simplify our life for sure.

I like the approach of directly appending to the right part.

backend/src/neuroagent/app/routers/tools.py
BoBer78: Not really, If I tried to remove it, FastAPI is not happy since it is a path param.

backend/src/neuroagent/app/dependencies.py
jankrepl: Was about the make the same comment. 

For debugging, you can actually set the `echo=True` when you create the SQLAlchemy engine (in `dependencies.py`) and then you will be able to see all SQL commands executed under the hood

```python
def setup_engine(
    settings: Settings, connection_string: str | None = None
) -> AsyncEngine | None:
    """Get the SQL engine."""
    if connection_string:
        engine_kwargs: dict[str, Any] = {
            "url": connection_string,
            "echo": True,  # Log all SQL queries
        }
        engine = create_async_engine(**engine_kwargs)
```

backend/src/neuroagent/agent_routine.py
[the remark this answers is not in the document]
jankrepl: My opinion is a bit more extreme...I would commit to the DB after each turn and not try to track all the changes in memory. I might be wrong but this will simplify the code.

backend/src/neuroagent/agent_routine.py
WonderPG: I don't remember exactly why we did it,  but there was a reason. Iirc it was mainly to unify them because each provider had a different way to build their uuid (e.g. openai it was call_xxx). Could we maybe conserve that logic or would it break something ?

backend/src/neuroagent/agent_routine.py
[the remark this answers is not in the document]
WonderPG: If this is what they advise, then great !

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
+                                ).model_dump(mode="json")
+                                args = json.dumps(validated_args)
+                            except ValidationError:
+                                args = input_args
+                            append_part(
+                                new_message, history, event.item, PartType.FUNCTION_CALL
+                            )
+                            # Tool call ready --> remove from tool_calls, add to tool_to_execute
+                            temp_stream_data["tool_calls"].pop(event.item.id, None)
+                            temp_stream_data["tool_to_execute"][event.item.id] = (
```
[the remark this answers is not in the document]
WonderPG: Sure we would have to wait for everything to be completed, but let's say we have 3 parallel tool calls:
- Tool 1 takes 10s to execute and has a small input body
- Tool 2 takes 1s to execute and has a large input body
- Tool 3 takes 1s to execute and has a large input body

Currently you would have to wait for all the tools to be defined (takes a couple seconds since 2 and 3 have a large input body) before they all start at the same time. Everything will finish 10s after they all started together. If each tool could start independently, you could cut a couple seconds from the global waiting time. Of course it is an edge case since this rarely happens, but it would come at no cost.

backend/src/neuroagent/tools/base_tool.py
WonderPG: Sorry I could've been more explicit. Some models require a different JSON Schema format (even for the same provider). Hence why we had redundant information 
```
"type": "function",
            "name": cls.name,
            "description": cls.description,
            "function": {
                "name": cls.name,
                "description": cls.description,
                "strict": False,
                "parameters": parameters,
            },
```
Some would take the info in the outer lever of nesting, some inside. Here you removed this duplicate hence my question.

backend/src/neuroagent/agent_routine.py
```diff
@@ -267,346 +288,317 @@ async def astream(
...
-                        for tool_call in message["tool_calls"]
-                    ]
-                else:
-                    tool_calls = []
+                        # Reasoning summary start
+                        case ResponseReasoningSummaryPartAddedEvent():
+                            temp_stream_data["reasoning"][event.item_id].summary.append(
+                                {"text": "", "type": "summary_text"}
+                            )
+                            yield f"data: {json.dumps({'type': 'start-step'})}\n\n"
```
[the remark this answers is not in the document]
BoBer78: Done in [37a3abb](https://github.com/openbraininstitute/neuroagent/pull/557/commits/37a3abb532f032e850aa09825a8fc3995b484181)

backend/src/neuroagent/app/app_utils.py
```diff
@@ -253,147 +238,84 @@ def format_messages_vercel(
...
+        tool_calls: dict[str, ToolCallPartVercel] = {}
+        metadata: dict[str, MetadataToolCallVercel] = {}
+
+        for part in msg.parts:
+            output = part.output or {}
+
+            if part.type == PartType.MESSAGE:
+                content = output.get("content")
+                if content and isinstance(content, list) and len(content) > 0:
+                    parts_data.append(TextPartVercel(text=content[0].get("text", "")))
```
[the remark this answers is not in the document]
BoBer78: `top_k` is no more supported in the Response API. I have no clue of why the content is a list ...

backend/src/neuroagent/app/app_utils.py
```diff
@@ -195,51 +198,33 @@ async def commit_messages(
...
-                {
-                    "tool_call_id": tc.tool_call_id,
-                    "name": tc.name,
-                    "arguments": tc.arguments,
-                    "validated": validation_status,
-                }
-            )
+            for item in content:
+                if item.get("type") == "text":
+                    parts_data.append({"type": "text", "text": item.get("text", "")})
```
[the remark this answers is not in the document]
BoBer78: Removed in [4b7fd49](https://github.com/openbraininstitute/neuroagent/pull/557/commits/4b7fd496a1f9837687cee7c8513ed33daee90db5)

backend/src/neuroagent/app/database/sql_schemas.py
BoBer78: I think this is all good, maybe it's my mistake

backend/src/neuroagent/app/database/sql_schemas.py
BoBer78: @jankrepl feel free to comment on this one.

backend/src/neuroagent/app/app_utils.py
```diff
@@ -578,15 +503,20 @@ async def filter_tools_and_model_by_conversation(
         # Send the OpenAI request
         model = "google/gemini-2.5-flash"
         start_request = time.time()
-        response = await openai_client.beta.chat.completions.parse(
-            messages=[{"role": "system", "content": system_prompt}, *openai_messages],  # type: ignore
+        response = await openai_client.responses.parse(
```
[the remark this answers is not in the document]
BoBer78: Done here : [49eea8f](https://github.com/openbraininstitute/neuroagent/pull/557/commits/49eea8ffa3eed2b8d2658eb08a3fa7b4700fdfc2)

frontend/src/components/chat/chat-input-inside-thread.tsx
```diff
@@ -89,7 +98,7 @@ export function ChatInputInsideThread({
             name="prompt"
             placeholder="Message the AI..."
             value={input}
-            onChange={handleInputChange}
+            onChange={(e) => handleInputChange(e.target.value)}
```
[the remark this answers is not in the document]
BoBer78: Good catch, fixed here : [ff2e292](https://github.com/openbraininstitute/neuroagent/pull/557/commits/ff2e292f417fff5110617a458be3cf0187470a0a)

backend/alembic/versions/25cefa8449c6_change_to_response_api.py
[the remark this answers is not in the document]
WonderPG: I am afraid that we will never remove it if we do it later because we will build on top of it. Do you think it has a real purpose ?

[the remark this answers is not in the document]
WonderPG: if so I am fine keeping it

backend/src/neuroagent/app/app_utils.py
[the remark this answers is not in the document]
BoBer78: I agree, I also tried asking LLMs ... it often mistakes the output beeing an array with a lot of possibilities, and the nested "content" array.
Let's try to dig deeper before we close this one, I agree. 
If @jankrepl you have any ideas, feel free to comment on this one.

backend/src/neuroagent/app/app_utils.py
```diff
@@ -253,147 +238,84 @@ def format_messages_vercel(
...
+                    tool_calls[tc_id].output = output.get("output") or "{}"
+                    metadata[tc_id].isComplete = part.is_complete
+
+        is_complete = all(part.is_complete for part in msg.parts) if msg.parts else True
+
+        msg_vercel = MessagesReadVercel(
+            id=msg.message_id,
+            role="user" if msg.entity == Entity.USER else "assistant",
+            createdAt=msg.creation_date,
+            isComplete=is_complete,
```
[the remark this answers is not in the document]
BoBer78: It has already been migrated. I think the message level `is_complete` is not necessary anymore

backend/src/neuroagent/tools/base_tool.py
BoBer78: It works now

backend/src/neuroagent/app/app_utils.py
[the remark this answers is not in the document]
BoBer78: Send a bug report to the OpenRouter team.

backend/alembic/versions/25cefa8449c6_change_to_response_api.py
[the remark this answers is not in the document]
BoBer78: My point is that, to debug : 
image
This is way simpler than to have to check manually in every output fields the type (or write some SQL statement). But sure, technically we can do this as a JSONB statement and the collumn does not have a "true" purpose.

PR #574, opened 2025-10-24
Update tool argument correctness evaluation

### LINKED ISSUE #560: `platform_news` gets 0 for argument correctness, why?
https://github.com/openbraininstitute/neuroagent/blob/ff3504c3ad13269874791092e8f26ecd632fe892/backend/eval/output/detailed.json#L2-L54

Potentially related to the logic that we strip out the defaults. But who knows. It is not impossible there is a bug in our code.

Also, feel free to consider dropping the argument correctness altogether in favor of https://github.com/openbraininstitute/neuroagent/pull/484#pullrequestreview-3361376574. However, I am not quite sure whether there is any added value to that

### LINKED ISSUE #573: Argument correctness gives <1 when it should give =1
See the `eval/output/detailed.json` ...it has happened in multiple cases.

The setup is that the expected and the actual tool calls are identical (-+ some tiny detail) but the argument correctness is saying the tool call should work differently.

### LINKED ISSUE #575: Remove useless level of nesting in `detailed.json`
The expected tool calls look like this

https://github.com/openbraininstitute/neuroagent/blob/918ae1f93a14eff1e78dabf8a3bd4df1b46d7df1/backend/eval/output/detailed.json#L5-L12

Whereas the actual tool calls look like this

https://github.com/openbraininstitute/neuroagent/blob/918ae1f93a14eff1e78dabf8a3bd4df1b46d7df1/backend/eval/output/detailed.json#L19-L28

TLDR - in the actual tool calls there is no need to have one extra layer of nesting (`tool_calls`)

The code is here: https://github.com/openbraininstitute/neuroagent/blob/918ae1f93a14eff1e78dabf8a3bd4df1b46d7df1/backend/src/neuroagent/scripts/evaluate_agent.py#L354

WonderPG:
Closes #573 closes #560 closes #575

### DIFF
--- a/backend/src/neuroagent/scripts/evaluate_agent.py
+++ b/backend/src/neuroagent/scripts/evaluate_agent.py
@@ -392,7 +392,7 @@ def collect_test_case_results(
         "params": test_case["params"].model_dump(),
         # Output data
         "ai_response": decoded_response["response"],
-        "actual_tool_calls": actual_tool_calls.model_dump(),
+        "actual_tool_calls": actual_tool_calls.model_dump()["tool_calls"],
         "results": eval_results.model_dump(),
     }
 


jankrepl (approved):
LGTM:)

PR #577, opened 2025-10-29
Context 7 - entitysdk and obione and neurom

### LINKED ISSUE #172: Code generator tool (only OBI packages)
**EDIT** 

We decided to go for Context 7 - https://context7.com/. Anyway, I keep all the description below if we ever want to do it ourselves from scratch.

# OLD DESCRIPTION

(This ticket is not about executing code - just generating it)

* I am very open on the choice of technologies that you go for. In the below I will just provide examples (pgvector + postgres).
* There are probably a lot of things online on this since popular tools like Cursor do something similar under the hood - one thing that is simpler in our case is that the libraries we want to support (NeuroM, BluePyEfe, ...) can be "pre-indexed" (no need to index them dynamically with each request)

Suggested workflow - RAG where the chunks=source files - for now I would only do it on `NeuroM` and `BluePyEfe` and `entitycore-sdk`

**Indexing**
0. Create a new table `code_embeddings` in postgres with the following schema +- (assuming you installed the `pg-vector` extension) + of course alembic migration script
```
repo: str
git_commit: str
file: str
embedding: vector(dim)
content: str <- so here you can dump the entire code - but in theory we can also retrieve it with git at any point during querying..maybe dumping it here is simpler
creation_date: datetime

```
1. Fix a git commit/tag of the repo
2. Traverse all relevant source code/documentation/markdown files (however, skip irrelevant files like `.gitignore`)
3. Embed each of them (using openai or ollama)
4. Dump embeddings and other fields into `code_embeddings`

**Querying**
1. Receive `query` from the user
2. Embed `query` (same embedding model as when indexing:D)
3. Approximate KNN in `code_embeddings` together with other filters (e.g. repo == "bluepyefe") to retrieve `top_k` source file
4. `system_prompt` and `query` and `top_k` source files content into an LLM and get a result

The querying will basically be a new tool

Potential issues
* The source files are huge
* No guarantee that the generated code is correct

jankrepl:
Closes #172 

# For reviewers
* I manually requested that both entitysdk and obione are tracked with context7 on their WWW: https://context7.com/
   * Not quite sure whether it will automatically check for new version or one needs to retrigger it manually but this is just a POC
* I did not provide any API_KEY since the API works without it - one just happens to have stricter rate limits


backend/src/neuroagent/rules/code-generation.mdc
```diff
@@ -0,0 +1,25 @@
+---
+description: Code generation
```
WonderPG: Imo we shouldn't have a one tool rule. This will blow out of proportion in the future.

WonderPG: Unless it is really impossible to make it work with the tool description

PR #591, opened 2025-11-03
Base class for excluded parameters + resolving of asset/hierarchy id …

### LINKED ISSUE #584: Exclude `brain_region__name` and similar from compatible EC tools
Many tools have a bunch of input parameters such as `brain_region__name`, `brain_region__name__in` etc... We prefer the LLM going for a resolution step using other tools, so banning these input parameters through
```
SkipJsonSchema[None] = Field(default=None, exclude=True)
```
would help.

Since a lot of input classes have to share this logic, it might make sense to create a parent "Exclusion" class in `base_tool.py` from which relevant tools would inherit from to keep the code tight. Feel free to extend it to other parameters that should generally be excluded such that the only way to use them is to perform a resolution step first.

### LINKED ISSUE #586: region -> hierarchy mapping done under the hood
# Status quo
A lot of current tools require the LLM to see and potentially override the  `within_brain_region_hierarchy_region_id`  parameter. In reality, the LLM never touches it since  we "hacked it" by hardcoding the identifier in the `default`. However, now that there are more hierarchies this approach is sketchy.

# How to fix it
Currently, the entitycore datamodel is the following
* hierarchy HAS MANY region
* region HAS ONE hierarchy

That means that we can simply imply the hierarchy id from the brain region id (the `GET /brain_region/$region_id` endpoint). And insert it manually inside `GET /some_entity` using the `within_brain_region_brain_region_id` and `within_brain_region_hierarchy_id` query params.

Also, this would mean that we need to hide `within_brain_region_hierarchy_region_id` from all the applicable tool inputs.

```python
    within_brain_region_hierarchy_id: SkipJsonSchema[None] = Field(default=None, exclude=True)
```

WARNING: In the future, the datamodel might change to be `MANY TO MANY`. But who knows.

# Example of the flow
1. `User`: Get me some morphologies from the thalamus using the hierarchy `$hierarchy_id` (note that here I don't mind if we use a default hierarchy if the user does not ask for a specific one)
2. `Tool("entitycore-brainregion-getall", hierarchy_id=$hierarchy_id, semantic_search="thalamus")`
3. `Tool("entitycore-cellmorphology-getall", brain_region=$region_id)` - HERE our tool would actually send two requests
   * `GET /brain_region/$region_id` that would give us the `hierarchy_id`
   * `GET /cell_morphollogy?within_brain_region_brain_region_id=$region_id&within_brain_region_hierarchy_id=$hierarchy_id`

# Discussions
* Sending two sequential requests in 1 tool is not necessarily the pattern we go for often
* Ideally, assuming the data model stays the same this should be done on the entitycore side of things and `within_brain_region_hierarchy_id` should be removed

WonderPG:
…within tools

Closes #584 closes #586

### DIFF
--- a/backend/src/neuroagent/rules/entitycore-tools.mdc
+++ b/backend/src/neuroagent/rules/entitycore-tools.mdc
@@ -3,9 +3,8 @@ description: Entitycore tools
 ---
 # ENTITYCORE TOOLS
 
-## Brain Region ID Resolution
+## Finding Entities within Brain Regions
 
-**MANDATORY workflow:**
 1. Resolve names to IDs using `entitycore-brainregion-getall` with `semantic_search` parameter
 2. Given the ID, retrieve some other entities located within that brain region using relevant tools (e.g., `entitycore-cellmorphology-getall`) and provide the parameter `within_brain_region_brain_region_id` with the resolved ID
 


backend/src/neuroagent/tools/base_tool.py
```diff
@@ -27,6 +28,29 @@ class EntitycoreMetadata(BaseMetadata):
...
+    brain_region__id__in: SkipJsonSchema[None] = Field(default=None, exclude=True)
+    brain_region__acronym: SkipJsonSchema[None] = Field(default=None, exclude=True)
+    brain_region__acronym__in: SkipJsonSchema[None] = Field(default=None, exclude=True)
+    within_brain_region_hierarchy_id: SkipJsonSchema[None] = Field(
+        default=None, exclude=True
+    )
+
+
+class EntitycoreExcludeNameParams(BaseModel):
+    """Exclude name parameters that the model should not use in name compatible EC endpoints."""
```
jankrepl: Here I would explicitly explain the purpose of this. We want to maximize the probability of the LLM using the `semantic_search`.

jankrepl:
What is a bit tricky is that entitycore allows for filtering that is "2 joins away" so for example 

```bash
src/neuroagent/tools/autogenerated_types/entitycore.py:4362:    morphology__brain_region__name: str | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:4365:    morphology__brain_region__name__in: list[str] | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:4368:    morphology__brain_region__name__ilike: str | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:4371:    morphology__brain_region__id: UUID | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:4374:    morphology__brain_region__id__in: list[UUID] | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:4377:    morphology__brain_region__acronym: str | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:4380:    morphology__brain_region__acronym__in: list[str] | None = Field(
```

or 

```
src/neuroagent/tools/autogenerated_types/entitycore.py:6983:    emodel__brain_region__name: str | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:6986:    emodel__brain_region__name__in: list[str] | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:6989:    emodel__brain_region__name__ilike: str | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:6992:    emodel__brain_region__id: UUID | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:6995:    emodel__brain_region__id__in: list[UUID] | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:6998:    emodel__brain_region__acronym: str | None = Field(
src/neuroagent/tools/autogenerated_types/entitycore.py:7001:    emodel__brain_region__acronym__in: list[str] | None = Field(
```

...

I think I will also disallow this one since it confuses the LLM. One can always just fetch the ids of brain regions satisfying some conditions via `entitycore-brainregion-getall` and then call the `entitycore-someentity-getall` tool X times for every brain region that satisfies that condition. It is not the most efficient thing but IMO it is a tradeoff.

backend/src/neuroagent/rules/entitycore-tools.mdc
```diff
@@ -7,13 +7,11 @@ description: Entitycore tools
 
 **MANDATORY workflow:**
```
BoBer78: Should we still call that "madatory workflow" ? maybe we should just call this section "Entities Handling"

BoBer78 (dismissed):
LGTM, one very minor comment about prompting. I tested it with most tools and it works great ! 
Thanks for the PR !

backend/src/neuroagent/rules/entitycore-tools.mdc
jankrepl: Great point. I did it here: a6a63ccc

PR #598, opened 2025-11-04
Yet another connectivity tool fix + version bump up

jankrepl:
I guess this one is again partially on me. I am just getting rid of all the temporary fixes I introduced in the tool that were supposed to bypass the bugs in the OBI API. Now the OBI API does not have bugs (just very complex logic) we can just go for the most vanilla implementation.


WonderPG (approved):
LGTM 🦫

PR #605, opened 2025-11-11
Push neuroagent to OBI private ERC

danifr:
for https://github.com/openbraininstitute/INFRA/issues/317


jankrepl (dismissed):
LGTM:) I guess once we make sure this works we could remove the pushing to the other ECR repositories in a separate PR:)

danifr:
> I guess once we make sure this works we could remove the pushing to the other ECR repositories in a separate PR:)

Yep exactly! for now let's just verify if this works. thanks Jan!!

PR #611, opened 2025-11-14
Introduce automatic model routing

### LINKED ISSUE #610: LLM tells you what it will do but does not actually do it
Example conversation

Image

I know we have observed this before . And I know that we hoped this rule would solve it

https://github.com/openbraininstitute/neuroagent/blob/9a269c63b610f3c16ee5876b1c6c825ee09d505a/backend/src/neuroagent/rules/tool-calling.mdc#L7

But yeh, let's try to prompt engineer things even more. I am not even sure if we can reproduce this/write an evaluation test case for it.

WonderPG:
Closes #610 

# Motivation
OpenAI claims to be doing the same. Based on the query, it routes it to a given model. We can get some great benefits from this imo. Cursor does it as well.

# Goal
Improve latency, cost and performance.

# How it works
I extended the tool selection method to do two conditional tasks
- Tool selection (if needed)
- Complexity assessment (if needed)

The complexity assessment is given for free (no extra latency, negligible extra cost) since the prompt is overall the same and we ask for one extra int as input.
The complexity is a general metric that can be used for multiple purpose. Here I use it for model/reasoning routing. In the future it can be extended to all sort of magic. Imo knowing in advance the query "complexity" is and will be a useful metric

# What we need to discuss
- Weather you like it or not. I understand that it can be unwanted.
- Which model/reasoning for which complexity. I went with some defaults for now that you might not like, I am definitely open to discussing which model/reasoning should be used for which complexity rating
- The way complexity is ranked. Currently it is on a scale of 0-10
- The prompt

# Alternatives
- Openrouter offers such possibility https://openrouter.ai/openrouter/auto. Not sure what it is worth but potentially better than what I did.
- Currently the models are restrained to OpenAI models. But nothing prevents from switching to openrouter models. In theory it should be a matter of replacing the string value in the routing function, it should be plug and play

This is a very open PR where the goal is to discuss the idea, feel free to criticize everything. Also imo it will integrate very well with #557 since this PR switches to response API, allowing to see reasoning summaries for extra smoothness.

### DIFF
--- a/backend/src/neuroagent/app/app_utils.py
+++ b/backend/src/neuroagent/app/app_utils.py
@@ -437,15 +437,13 @@ def complexity_to_model_and_reasoning(complexity: int) -> dict[str, str | None]:
     if complexity <= 1:
         return {"model": "openai/gpt-5-nano", "reasoning": "minimal"}
     elif complexity <= 3:
-        return {"model": "openai/gpt-4.1-mini", "reasoning": None}
+        return {"model": "openai/gpt-5-nano", "reasoning": "low"}
     elif complexity <= 5:
         return {"model": "openai/gpt-5-mini", "reasoning": "low"}
-    elif complexity <= 7:
+    elif complexity <= 8:
         return {"model": "openai/gpt-5-mini", "reasoning": "medium"}
-    elif complexity <= 9:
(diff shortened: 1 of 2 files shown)


jankrepl (dismissed):
LGTM:)

PR #617, opened 2025-11-18
Exa standalone tool

### LINKED ISSUE #608: Replace exa mcp server with custom tools with direct API calls
# What?
* Literature search
* Web search
* Crawling

# Why?
* We are hardcoding an old version of the mcp server since they are moving away from our desired way of doing things
https://github.com/openbraininstitute/neuroagent/blob/9a269c63b610f3c16ee5876b1c6c825ee09d505a/backend/src/neuroagent/mcp.json#L35
* The tool output contains a lot of information the final user should not even see/care about. This is especially relevant since this ticket might be soon addressed: https://github.com/openbraininstitute/prod-explore-functionality/issues/385

# Comments
* Keep the tool name + tool slug exactly the same..this way it will be easier to create statistics about tool usage
* This will require changing the name of the env variable secret

WonderPG:
Closes #608 

I took the freedom to make minimal changes to the tools compared to their original description as MCP. Let me know if you would remove some

## Literature Search
- Added possibility to filter by `start_published_date` and `end_published_date`
- Explicitly restricted to at most one image per result in the request
- Added `"category": "research paper"` query param (which might have broken cloudflare globally)
- Added the following domains from which we can get results alongside with the ones that were already included in the MCP `"nature.com","cell.com", "elsevier.com", "elifesciences.org", "frontiersin.org"`
- Increased `maxCharacters` to 3000 instead of 2000, i.e. excerpts are 1000 character longer. I did this because a lot of time 2000 characters included only the page metadata and didn't reach the actual relevant content

## Web Search
- Added possibility to filter by `start_published_date` and `end_published_date`
- Increased `maxCharacters` to 3000 instead of 2000, i.e. excerpts are 1000 character longer
- Explicitly restricted to at most one image per result in the request.

## Read Paper
- Made it compatible with batch requests since the API allows for it. The MCP accepted only one url at a time
- Explicitly restricted to at most **3** image per result in the request.

### DIFF
--- a/backend/src/neuroagent/tools/literature_search.py
+++ b/backend/src/neuroagent/tools/literature_search.py
@@ -64,6 +64,18 @@ class LiteratureSearchTool(BaseTool):
         "Find literature on this topic",
         "Look up academic papers",
         "Search for research papers",
+        "Find papers about",
+        "Search scientific literature",
+        "Look for studies on",
+        "Find research about",
+        "Search for publications",
+        "What papers exist on",
+        "Find academic articles",
+        "Search for scholarly articles",
+        "Look up scientific papers",
+        "Find recent papers on",
+        "Search neuroscience literature",
+        "What research has been done on",
     ]
     description: ClassVar[str] = (
         "Search across 100M+ research papers with full text access using Exa AI - performs targeted academic paper searches with deep research content coverage. "
--- a/backend/src/neuroagent/tools/read_paper.py
+++ b/backend/src/neuroagent/tools/read_paper.py
@@ -99,7 +99,7 @@ class ReadPaperTool(BaseTool):
 
             if response.status_code != 200:
                 raise ValueError(
-                    f"The Exa search endpoint returned a non 200 response code. Error: {response.text}"
+                    f"The Exa contents endpoint returned a non 200 response code. Error: {response.text}"
                 )
             data = response.json()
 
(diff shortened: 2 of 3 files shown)


backend/src/neuroagent/tools/read_paper.py
```diff
@@ -0,0 +1,116 @@
...
+                headers={
+                    "x-api-key": self.metadata.exa_api_key,
+                    "Content-Type": "application/json",
+                },
+                json=payload,
+            )
+
+            if response.status_code != 200:
+                raise ValueError(
+                    f"The Exa search endpoint returned a non 200 response code. Error: {response.text}"
```
BoBer78: I guess it is not the "search" endpoint.

BoBer78 (approved):
A couple of small comments, other than that LGTM !

backend/src/neuroagent/tools/read_paper.py
WonderPG: Done in 8d16b70c

backend/src/neuroagent/tools/literature_search.py
```diff
@@ -0,0 +1,149 @@
...
+                "elsevier.com",
+                "elifesciences.org",
+                "frontiersin.org",
+            ],
+            "contents": {
+                "text": {
+                    "maxCharacters": 3000  # Len of excerpts. Not specified = full page
+                },
+                "livecrawl": "preferred",
+                "extras": {"imageLinks": 1},
```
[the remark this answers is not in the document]
BoBer78: Same, I tested a bit, only got one every time.

PR #633, opened 2025-11-25
Use  in autogen (not for obi-one)

### LINKED ISSUE #630: Prevent autogen from turning shared Literal into RootModel
From #629 a comment from @WonderPG 

> I had a quick investigation, there is this flag --collapse-root-models that is meant to remove the root models. Feel free to give it a go. Please test other tools after this change because I remember quickly testing it at some point and some changes in other tools were needed. If it's too annoying I'm fine with the current solution.

WonderPG:
Closes #630


BoBer78:
I am fine if we get rid of it.

WonderPG:
Sounds good

PR #637, opened 2025-11-26
suppress warnings in python tool

### LINKED ISSUE #634: Warnings in `run-pyton` executed code are treated as errors
## Description
When running a code using the `run-python` tool, it can happen that it raises a warning. It seems like the warning is caught in the stderr which leads the tool to return an error:
```
{
  "result": {
    "status": "error",
    "error_type": "install-error",
    "error": "<exec>:216: DeprecationWarning:\n\n*scattermapbox* is deprecated! Use *scattermap* instead. Learn more at: https://plotly.com/python/mapbox-to-maplibre/\n\n<exec>:219: DeprecationWarning:\n\n*scattermapbox* is deprecated! Use *scattermap* instead. Learn more at: https://plotly.com/python/mapbox-to-maplibre/\n\n"
  },
  "storage_id": []
}
```
Warnings should be allowed to go through

WonderPG:
Closes #634


WonderPG:
They do because the warnings go into `stderr` and we have 
```
# Check for execution errors
            if stderr:
                return FailureOutput(error_type="install-error", error=stderr.decode())
```
before returning anything else

PR #647, opened 2025-12-01
Use gpt-5-nano as suggestion/title model

### LINKED ISSUE #639: Check if we can go for `gpt-5-nano` for "questions suggestions" and "title generation"
## Check if feasible 

Check if the model is good enough to do these tasks. Also it should not be significantly slower. 

## Comparison : 

Intelligence:
Image

Price: 
Image

Speed:
Image

WonderPG:
Closes #639


BoBer78 (approved):
LGTM

PR #669, opened 2026-01-16
Initialize documentation

### LINKED ISSUE #666: Initialize docs + 2 github actions
Do it exactly like it was done in other repos (e.g. https://github.com/openbraininstitute/core-web-app/pull/1265/files)

However, make sure to not only add the "Check docs update" action but also on new commits in `main` trigger a "Push to sanity action"

The latter action should be a bit clever
* Let's use the `{repo}/{markdown_name}` as a unique identifier
* Check for modifications (md5 hash?) before actually pushing
* Only trigger the action if `/docs/*` had a change

Note that maybe it would make more sense to make both of the actions "Open" action so that they can be reused across all the repos

jankrepl:
Closes #666

# TODO
- [x] Setup mkdocs
- [x] Fix autogen
- [x] Initialie the docs
- [x] Write check docs action + "skip docs" tag
- [ ] ~Write deploy docs action (only on `main`)~ - actually, I decided to put it in the `sanity-docs` repo for now.
- [x] Use optional dependencies instead of dependency groups
- [x] Fix autogen issue

### DIFF
--- a/backend/src/neuroagent/tools/obi_expert.py
+++ b/backend/src/neuroagent/tools/obi_expert.py
@@ -382,7 +382,9 @@ class DocumentationGithub(SanityDocument):
     title: str | None = None
     slug: dict[str, Any] | None = None
     keywords: list[str] | None = None
-    product: str | None = None
+    product: list[str] | None = None
+    repository: str | None = None
+    commit: str | None = None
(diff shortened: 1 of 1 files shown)


BoBer78 (approved):
LGTM ! ;)

PR #672, opened 2026-01-21
Fix DB mesage loading (duplicate text part with parallel tool calls)

### LINKED ISSUE #649: Duplicate "message before tool" when loading message from DB (with parallel tool calls)
## Explanation

This only happens when there are parallel tool calls. 
So when streaming, there is no issue : 

Image

But when you load back a message from the Database, you get the message from the AI repeated. Since there is some fancy caching in the frontend, you have to refresh to see it. Here is a small video : 

https://github.com/user-attachments/assets/660617ed-a753-4aa8-8f5a-ccc9c9e603b1

We have the same issue in our frontend. It probably comes from the infamous `message_to_vercel_format`. 

## Reproduce the bug

Ask for 3 morphologies in the thalamus. Then refresh the page. Check the parallel thumbnail generation.

BoBer78:
Closes #649

### DIFF
--- a/backend/src/neuroagent/tools/autogenerated_types/entitycore.py
+++ b/backend/src/neuroagent/tools/autogenerated_types/entitycore.py
@@ -83,7 +83,7 @@ class AnalysisNotebookResultUpdate(BaseModel):
 
 
 class CountMax(RootModel[int]):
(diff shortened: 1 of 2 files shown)


WonderPG (approved):
It is fixed. Thanks 🦫

PR #683, opened 2026-01-30
Write exclusively to tmp

### LINKED ISSUE #678: Don't write anywhere else than `/tmp`
Our backend seems to be writing to disk at multiple places. As long as it is `/tmp` it is fine. But it actually writes to different places too.

One example is related to the mcp servers OR the python tool because they use npm .

```bash
npm ERR! Log files were not written due to an error writing to the directory: /root/.npm/_logs
```

Note that there might be other places. 

The easiest way to be sure it works is to run the backend via a docker container with the following flags

```bash
 docker run --read-only --tmpfs /tmp your-image
```

ALTERNATIVE APPROACH
You enumerate all the places where the backend writes and then we add temp volumes to all of them.

WonderPG:
Closes #678 

@danifr In the `docker-compose.yml` I had to add the following lines to reproduce your setup
```
read_only: true
tmpfs:
    - /tmp:exec,mode=1777
```
such that I could execute files from the `/tmp` folder. Is it something you can replicate in Terraform ?

### DIFF
--- a/backend/src/neuroagent/tools/autogenerated_types/obione.py
+++ b/backend/src/neuroagent/tools/autogenerated_types/obione.py
@@ -608,6 +608,18 @@ class InhibitoryNeurons(BaseModel):
     )
 
 
+class Initialize2(BaseModel):
+    model_config = ConfigDict(
+        extra='ignore',
+    )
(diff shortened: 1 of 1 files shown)


jankrepl (approved):
I guess LGTM:)

I guess the real test is to see whether it works on ECS:)

PR #687, opened 2026-02-02
Pricing info in OBI Expert

### LINKED ISSUE #642: Add everything related to pricing in obi expert
There are a bunch of things that go into it:

Image

Maybe this one would use an extra wrapper on our side for the LLM to understand it. Please experiment

jankrepl:
Closes #642 

- [x] Add a new document type
- [x] Add unit test that holds an actual response body from sanity
- [x] Add one or multiple deepeval test cases


WonderPG:
I tried your deepeval question (difference between free and pro plan) multiple times and it kept going back and forth between `documentationProduct` and `page` document type, which did not work.  I know this is how Sanity is organized, but wouldn't it make sense to ditch this filter and search across all document types ? They are very often the reason why the search fails.

Also with semantic search (hopefully), the most relevant document should come up no matter the category.

PR #690, opened 2026-02-02
Simplify docker compose

### LINKED ISSUE #686: Simplify initial setup (docker compose, environment files,...)
I suggest we copy paste the logic of other OBI repos (e.g. entitycore).

Namely

* Makefile (the user only runs `make blabla` and they don't have to worry about the logic under the hood)
* Deal with the service references in docker compose in the most standard way (research it) .. e.g. `localhost:5432` vs `db:5432` - ideally without having to have too env/variables files lying around
* Local commands (frontend + backend are running outside docker) vs docker commands (everything inside)
* No extra `.env` files inside `backend/` or `frontend/
* ...

Currently, external people struggle to run neuroagent locally. Some reasons why

* The README.md makes you believe there are only like 5 env vars. But in reality to be able to run the tools there are way more

WonderPG:
Closes #686 

I went with a very simply approach:
- Provide the `.env` and `.env.local` files that one must have in its local tree
- Remove all of the env vars in `environment` except for the ones that define elements that are being managed by docker compose (e.g. db credentials, backend port, minio url etc...)

`environment` takes precedence over `env_file` so we override variables that would change between local/docker-compose directly in `docker-compose.yml`

Please test by running `docker-compose up` and using our frontend. It should work without changing anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,22 +9,22 @@ LLM agent for interacting with neuroscience tools. Enables ChatGPT-style interac
 The easiest way to run the project is with Docker Compose:
 
 1. Install Docker and Docker Compose
-2. Create `.env` file with required variables:
+2. Create a `backend/.env` file with required variables. Configurable variables are listed in `backend/.env.example`. Please contact the Machine Learning team to get the secret values.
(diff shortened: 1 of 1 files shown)


jankrepl:
Can you also modify the README @WonderPG? Because it instructs the user to use `.env`

PR #691, opened 2026-02-02
Fix ilike usage by the LLM

### LINKED ISSUE #671: LLM uses `*` as wildcard for `ilike` query params
Note sure how reproducible it is, but the LLM called the get all circuits with the following params

```json
{
  "name__ilike": "rCA1*",
  "scale": "small"
}
```
Note that the `*` is not the correct wildcard to use (it should either not put a wildcard or use `%`).

Make sure we did not actually give it the wrong idea somewhere in our rules/descriptions.

And maybe even override the description of the applicable query params (actually, don't like this idea since it would require probably hundreds of overrides unless we can put it in one of the shared classes)

WonderPG:
Closes #671

### DIFF
--- a/backend/src/neuroagent/tools/autogenerated_types/obione.py
+++ b/backend/src/neuroagent/tools/autogenerated_types/obione.py
@@ -608,6 +608,18 @@ class InhibitoryNeurons(BaseModel):
     )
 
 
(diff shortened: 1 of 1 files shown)


jankrepl (approved):
LGTM:) Thanks

PR #692, opened 2026-02-02
Info tool

### LINKED ISSUE #695: Simulation config info tool
We have a tool that generates a valid simulation form JSON.

```
TEXT -> JSON
```

We were requested to also do something slightly different - let the user ask questions about the simulation configuration json schema. 

```
TEXT -> TEXT
```

I guess the idea here would be json `model_json_schema()` in to the prompt together with the question.

james-isbister:
Closes #695


jankrepl:
@james-isbister feel free to check it:)

Don't worry about the CI for now. Also, the base branch is `state_sharing` so this would have to be adjusted a bit to be mergable into `main`. But yeh, feel free to first determine whether you want a tool like this. I will do the rest.

PR #697, opened 2026-02-06
Old questions test cases

### LINKED ISSUE #693: Past questions -> Deepeval test cases
Use our internal DB to generate new test cases.

Now, there are 2 things you need to define

* expected tool calls - make an assumption that we use `obi-expert` with `documentationProduct`
* expected answer - Here you either need to ask a scientist or find it in the existing docs

Let's try to do 50 of them in the first iteration

jankrepl:
Closes #693 

# For reviewers
* There are 59 new test cases
* By default none of the test cases are run. To actually run them you need to set the `--exclude-tags=""`
* I added the `-n / --dry-run` flag to the CLI to just enumerate the tests that would be run
* All the test cases I added have the `benchmark` and `$THREAD_ID` tags and other tags based on what the `user.md` is
* Once we merge this PR the idea would be to assign different tags to different people and create new issues

# TODO
- [x] Create test cases
  - PLACHOLDER in `expected_output.md`
  - {} in `expected_tool_calls.json`
  - `"benchmark", `$thread_id` in `params.json`
- [x] Make sure that test cases that have `"benchmark"` tag are not run in the CI
- [x] Add dry-run functionality to the CLI

### DIFF
--- a/backend/src/neuroagent/scripts/evaluate_agent.py
+++ b/backend/src/neuroagent/scripts/evaluate_agent.py
@@ -567,7 +567,9 @@ async def eval_sample(
             if response.status_code == 200:
                 thread_id = response.json()["thread_id"]
             else:
-                raise RuntimeError(f"Failed to create a thread. Reason: {response.text}")
+                raise RuntimeError(
+                    f"Failed to create a thread. Reason: {response.text}"
+                )
             try:
                 # Build request body with defaults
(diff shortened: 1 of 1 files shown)


backend/eval/input/aging_vs_tbi_hippocampus/params.json
```diff
@@ -0,0 +1,3 @@
+{
+  "tags": ["benchmark", "b121f6d3-5ec1-475b-b28d-976e9a5cd0b5", "simulation", "hippocampus", "aging", "tbi", "mouse", "comparison"]
```
WonderPG: Is having a UUID in the tag expected ?

PR #706, opened 2026-02-12
nodeset failure doesn't raise an error anymore in circuit simulation …

### LINKED ISSUE #705: Sim config: Do not raise exception if there are no predefined nodesets
# Current behavior
Our current logic is that if there are no predefined nodesets for a given circuit, we raise an exception that looks like this

```
The nodesets endpoint returned a non 200 response code. Error: Internal Server Error
```

It corresponds to this block of code

https://github.com/openbraininstitute/neuroagent/blob/d625290629d0bf5004e45963d85ed0c66c76b359/backend/src/neuroagent/tools/obione_generatesimulationsconfig.py#L163-L171

# What should we do instead?
Now, after discussing with @g-bar we concluded that this is a bit too harsh. Predefined nodesets (neuronsets) are one of MANY ways how one can specify sets of neurons. See below all the options shown nicely in the UI

Image

We should instead just silently ignore the issue (maybe set the node sets to an empty list or None) and then dynamically change the inner LLM prompt and mention that we did not find any predefined nodesets and that it cannot use them.

# Further comments

* For testing in staging/dev feel free to use the following circuit that does not have predefined nodesets: `c940188a-5408-4185-ab3b-dbb20de16621`
* If you think this is a data issue and not a neuroagent issue (e.g. the nodesets should always return something) then you can ping some scientists and try to get their take

WonderPG:
Closes #705

### DIFF
--- a/backend/src/neuroagent/app/dependencies.py
+++ b/backend/src/neuroagent/app/dependencies.py
@@ -659,11 +659,7 @@ Current time: {datetime.now(timezone.utc).isoformat()}"""
         body = await request.json()
         if body.get("frontend_url"):
             system_prompt += f"""
-Information extracted from the user's current page URL: {extract_frontend_context(body["frontend_url"]).model_dump()}.
-NOTE: This context contains only IDs (e.g., brain_region_id, current_entity_id), not names or labels. Use these IDs ONLY when:
-- The user clearly references "this brain region," "current page," or similar terms
-- The user's query unmistakably concerns the viewed entity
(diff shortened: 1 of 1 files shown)


backend/src/neuroagent/app/config.py
```diff
@@ -290,7 +290,13 @@ def parse_mcp_servers(cls, data: Any) -> Any:
...
         mcps: dict[str, dict[str, Any]] = {"servers": {}}
         # Parse and set the mcp servers
-        for server, config in json.loads(servers).items():
+        try:
+            parsed_servers = json.loads(servers)
+        except Exception as e:
+            logger.warning(f"Could not parse mcp.json: {e}")
+            return data
+
+        for server, config in parsed_servers.items():
```
WonderPG: This allows you to comment out the `mcp.json` file for your backend to start faster (bypassing context7)

PR #712, opened 2026-02-13
Sim generation: Add to Info, Recording + PredefinedNeuronSet prompts

james-isbister:
- Minor suggested updates to sim generation tool system_prompt

### DIFF
--- a/backend/src/neuroagent/tools/obione_generatesimulationsconfig.py
+++ b/backend/src/neuroagent/tools/obione_generatesimulationsconfig.py
@@ -124,6 +124,10 @@ Call this tool whenever users:
 
 # Verbosity
 - Provide concise, high-level summaries clearly stating the overall setup and nature of each simulation.
+
+# Future features (currently unsupported)
+- Users might still ask for particular features which have not yet been implemented. If they are listed here, you can tell them they are planned for future.
+- In future, the user will be able to record voltage from any location in neurons, and also specify extracellular recordings.
 """
     description_frontend: ClassVar[
         str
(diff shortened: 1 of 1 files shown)


backend/src/neuroagent/tools/obione_generatesimulationsconfig.py
```diff
@@ -213,7 +223,8 @@ async def arun(self) -> GenerateSimulationConfigOutput:
   - `block_name`: the dictionary key (e.g., "all_neurons")
   - `block_dict_name`: the parent dictionary name (e.g., "neuron_sets", "timestamps")
 - Avoid duplicate or unused neuron sets unless specifically requested
-{f"- Whenever you see a `circuit_property_type = CircuitNodeSet` in the schema of the neuron_sets you want to use, the string value within `node_set` must be one of the following: {nodesets}" if nodesets is not None else ""}
+- {f"Whenever you see a `circuit_property_type = CircuitNodeSet` in the schema of the neuron_sets you want to use, the string value within `node_set` must be one of the following: {nodesets}" if nodesets is not None else ""}
+- Usage of PredefinedNeuronSets should be avoided where possible. Prefer using other neuron set types.
```
jankrepl: Interesting. Why should they be avoided? Maybe we should not support them in the first place?

james-isbister: They're still useful for using very specific previously defined NeuronSets (such as spatially defined hexagonal columns which aren't yet supported through other NeuronSets)

We want to move towards everything being parametrically defined as a NeuronSet though

PR #715, opened 2026-02-16
Various simulation config tickets

### LINKED ISSUE #698: Add custom frontend url + state for each test case
Right now, we hardcode the frontend url in https://github.com/openbraininstitute/neuroagent/blob/5a7d83a710b315d103c7ef8f34b57846872432ea/backend/src/neuroagent/scripts/evaluate_agent.py#L460

However, we should make this dynamic and make it a part of the `params.json` (under key `frontend_url`) and make sure the `evaluate_agent.py` honors that.

Note that I am fine with having some reasonable default if that given test does not depend on the frontend url - in that case `params.json` does not have to have that key? 

# Why does it matter?
Well if the user asks "What do I see on the screen?" the correct answer will depend on where they are.

### LINKED ISSUE #707: Renaming: generate -> design
Currently, in  `obione_generatesimulationsconfig.py` we use the word `generate` a lot. 

However, that is very confusing because in the UI the word generate is already present and it corresponds to "storing the configuration so that it can used for simulation/retrieved later" - or in other words, it stores things in entitycore.

Image

Let's just stop using the word "generate" completely and use "design" instead

Once this ticket is done, we need to also update CWA

### LINKED ISSUE #710: Explain relationship between simulation campaigns and design configuration tools
I don't think our agent understands how the two are related (feel free to prove me wrong).

Simulation campaigns are simulation configurations that were saved in entitycore basically once the `Generate simulation` button was clicked.

Whereas the design tools only help initialize the simulation configs (see #708 and #709) but it is only up to the user to accept them/modify them in the interface and only then save (Generate button).

Also note that currently the simulation campaigns are immutable and one cannot modify them. IMO the agent should suggest duplicating an existing simulation campaign (that has a UUID) so that they can "modify an existing campaign" exactly like the UI suggests.

Maybe we should also explain how the other simulation-generation,simulation-execution,simulation-result are related?

### LINKED ISSUE #711: Get state tool
This issue is motivated by a real problem @ayimaok hit

1. Go to past simulation campaigns for microcircuits 
2. (I just learned this) The frontend will cleverly insert the simulation config in the state
3. Open chat and ask it to describe what happened
4. The Agent LLM has no idea what is inside the state.

Note that when one creates a new simulation config from scratch this issue is not there because the tool output contains the whole state. So the Agent LLM does know what is inside.

Also, let's dump the whole state in the output of this new get state tool (we already do it in the generate sim config tool so whatever)

jankrepl:
Closes #711, closes #710 and closes #707 and closes #698

TODO
- [x] Rename: generate -> design
- [x] Get state tool
- [x] Explain relationship between simulation design vs the entitycore simulation tools
- [x] Allow passing `shared_state` and `frontend_url` via `params.json` optionally
- [x] Adjust old deepeval test cases
- [x] Add new deepeval test cases


backend/eval/input/obione_designcircuitsimulationscanconfig/expected_tool_calls.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "name": "obione-generatesimulationsconfig",
+    "name": "obione-designcircuitsimulationscanconfig",
```
WonderPG: So the new name contains `scan` as well ? Isn't it a bit confusing ?

WonderPG (commented):
Thanks for the PR !

backend/eval/input/obione_designcircuitsimulationscanconfig/expected_tool_calls.json
jankrepl: Maybe but I just wanted to follow the obi-one naming.

WonderPG: But a scan is a set of simulation. Here we generate one no ?

WonderPG: Fair

WonderPG (dismissed):

PR #737, opened 2026-03-23
Navigator tool

WonderPG:
Closes #724


backend/src/neuroagent/app/schemas.py
```diff
@@ -308,8 +308,6 @@ class SearchMessagesList(BaseModel):
...
     """Output of the Context Analyzer tool."""
 
-    raw_path: str
-    query_params: dict[str, list[str]]
-    brain_region_id: UUID | None = None
-    observed_entity_type: str | None = None
-    current_entity_id: UUID | None = None
+    route_description: str
+    path_params: dict[str, str]
+    search_params: dict[str, str]
```
jankrepl: Just a thought...Probably an overkill

The sitemap has descriptions for all the path and search params...so I am wondering whether we should not do something like 

`dict[str, tuple[str, str]]`   (the first entry is the value and the second one is the description)


But probably that is an overkill and the LLM will get the "context" just by reading the param name + value

WonderPG: So something like `{"br_id": ("1929457-fd12-...", "Current brain region")}` ? It is a bit more pythonic I agree. Or we could do `list[dict[str, str]]` with something like `[{"name": "br_id", "description", "Current brain region", "value": "1929457-fd12-..."}, {"name": "hr_id"...}]` to make it even more explicit. What do you think ?

jankrepl: Yeh, `list[dict[str, str]]` sounds good:) But then maybe we can have a `BaseModel` for it

```python
class Param(BaseModel):
    name: str
    value: Any
    description: str
```

and do `path_params: list[Param]` and `search_params: list[Param]`

WonderPG: Done in [fec2a11](https://github.com/openbraininstitute/neuroagent/pull/737/commits/fec2a11161478053c8b7e6c3bbfb65542b2d8926)

PR #741, opened 2026-03-30
Run deepeval in the CI on demand

jankrepl:


WonderPG (approved):
Not claiming I understand the code, but I have been able to download and open the report. Thanks !

PR #744, opened 2026-04-13
Another deepeval benchmark run

jankrepl:
- [x] Make sure the model is encouraged to search through docs without user's permission
- [x] Expect virtual lab expected tool calls - like virtual all of them should use the obi-expert
- [x] Maybe make sure we don't do queryless searches? 
```
[
  {
    "name": "obi-expert",
    "arguments": {
      "document_type": "documentationProduct",
      "page_size": 10
    }
  }
]
```


BoBer78 (approved):
LGTM !

